PR #20, opened 2020-08-24
Validate allowedComponents

henrytao-me:
This PR adds error handling for unsupported components by throwing exception. It also restricts components that are created by `root.createComponent` or `root.createText` which means defining components as json object isn't allowed.

Ref https://github.com/Shopify/app-extension-libs/issues/802

## How does it work?
- Check for allowedComponents when calling root.createComponent or root.createText.
- Have a weakset to cache all components created by remote root. 

## How to tophat?

- In https://github.com/shopify/app-extension-libs, do `dev up`
- In this PR, do `yarn && yarn build`
- In packages, copy: core, react, rpc, types, web-workers from this PR to `app-extension-libs` repo at `./node_modules/@remote-ui`
- In `app-extensions-lib`, run `dev s`
- Change `worker.render` to only support `Card` component.
```ts
  useEffect(() => {
    (async () => {
      const componentsList = Object.keys(components);
      if (!script || !componentsList.length) {
        return;
      }
      await worker.render(extensionPoint, api, ['Card'], (type, ...args) => {
        // Have a proper fix in remote-ui core library later
        // Ref: https://github.com/Shopify/app-extension-libs/issues/436#issuecomment-622008563
        retain(args);

        return receiver.receive(type, ...args);
      });
    })();
  }, [components, extensionPoint, api, receiver, script, worker]);
```
- Open https://argo-playground.myshopify.io/
- Verify that console log is output as below.

image

### DIFF
--- a/packages/core/src/receiver.ts
+++ b/packages/core/src/receiver.ts
@@ -5,7 +5,6 @@ import {
   ACTION_REMOVE_CHILD,
   ACTION_UPDATE_PROPS,
   ACTION_UPDATE_TEXT,
-  ACTION_ERROR,
 } from './types';
 import type {
   ActionArgumentMap,
@@ -33,7 +32,6 @@ interface RemoteChannelRunner {
   removeChild(...args: ActionArgumentMap[typeof ACTION_INSERT_CHILD]): void;
   updateProps(...args: ActionArgumentMap[typeof ACTION_UPDATE_PROPS]): void;
   updateText(...args: ActionArgumentMap[typeof ACTION_UPDATE_TEXT]): void;
-  error(...args: ActionArgumentMap[typeof ACTION_ERROR]): void;
 }
 
 export function createRemoteChannel({
@@ -42,7 +40,6 @@ export function createRemoteChannel({
   removeChild,
   updateProps,
   updateText,
-  error,
 }: RemoteChannelRunner): RemoteChannel {
   const messageMap = new Map<number, Function>([
     [ACTION_MOUNT, mount],
@@ -50,7 +47,6 @@ export function createRemoteChannel({
     [ACTION_INSERT_CHILD, insertChild],
     [ACTION_UPDATE_PROPS, updateProps],
     [ACTION_UPDATE_TEXT, updateText],
-    [ACTION_ERROR, error],
   ]);
 
   return (type, ...args) => (messageMap.get(type) as any)(...args);
@@ -119,9 +115,6 @@ export class RemoteReceiver {
       const text = this.attached.get(id) as RemoteTextSerialization;
       text.text = newText;
       this.enqueueUpdate(text);
-    },
-    error: () => {
-      // Noop
     },
   });
 
--- a/packages/core/src/receiver.ts
+++ b/packages/core/src/receiver.ts
@@ -5,6 +5,7 @@ import {
   ACTION_REMOVE_CHILD,
   ACTION_UPDATE_PROPS,
   ACTION_UPDATE_TEXT,
+  ACTION_ERROR,
 } from './types';
 import type {
   ActionArgumentMap,
@@ -32,6 +33,7 @@ interface RemoteChannelRunner {
   removeChild(...args: ActionArgumentMap[typeof ACTION_INSERT_CHILD]): void;
   updateProps(...args: ActionArgumentMap[typeof ACTION_UPDATE_PROPS]): void;
   updateText(...args: ActionArgumentMap[typeof ACTION_UPDATE_TEXT]): void;
+  error(...args: ActionArgumentMap[typeof ACTION_ERROR]): void;
 }
 
 export function createRemoteChannel({
@@ -40,6 +42,7 @@ export function createRemoteChannel({
   removeChild,
   updateProps,
   updateText,
+  error,
 }: RemoteChannelRunner): RemoteChannel {
   const messageMap = new Map<number, Function>([
     [ACTION_MOUNT, mount],
@@ -47,6 +50,7 @@ export function createRemoteChannel({
     [ACTION_INSERT_CHILD, insertChild],
     [ACTION_UPDATE_PROPS, updateProps],
     [ACTION_UPDATE_TEXT, updateText],
+    [ACTION_ERROR, error],
   ]);
 
   return (type, ...args) => (messageMap.get(type) as any)(...args);
@@ -115,6 +119,9 @@ export class RemoteReceiver {
       const text = this.attached.get(id) as RemoteTextSerialization;
       text.text = newText;
       this.enqueueUpdate(text);
+    },
+    error: () => {
+      // Noop
     },
   });
 
--- a/packages/core/src/root.ts
+++ b/packages/core/src/root.ts
@@ -7,8 +7,6 @@ import {
   ACTION_UPDATE_TEXT,
   KIND_COMPONENT,
   KIND_TEXT,
-  ACTION_ERROR,
-  ERROR_UNSUPPORTED_COMPONENT,
 } from './types';
 import type {
   Serialized,
@@ -46,6 +44,7 @@ export function createRemoteRoot<
   const props = new WeakMap<Component, any>();
   const texts = new WeakMap<Text, string>();
   const tops = new WeakMap<CanBeChild, HasChildren>();
+  const componentsSet = new WeakSet<CanBeChild>();
 
   let currentId = 0;
   let mounted = false;
@@ -55,6 +54,10 @@ export function createRemoteRoot<
       return children.get(remoteRoot) as any;
     },
     createComponent(type, ...rest) {
+      if (!validateAllowedComponent(type, components)) {
+        throw new Error(`Unsupported component: ${type}`);
+      }
+
       let initialProps = rest[0];
       const initialChildren = rest[1];
 
@@ -129,9 +132,15 @@ export function createRemoteRoot<
         children.set(component, strict ? Object.freeze([]) : []);
       }
 
+      componentsSet.add(component);
+
       return (component as unknown) as RemoteComponent<typeof type, Root>;
     },
     createText(content = '') {
+      if (!validateAllowedComponent('Text' as any, components)) {
+        throw new Error(`Unsupported component: Text`);
+      }
+
       const id = `${currentId++}`;
 
       const text: RemoteText<Root> = {
@@ -148,6 +157,8 @@ export function createRemoteRoot<
       makePartOfTree(text);
       makeRemote(text, id, remoteRoot);
       texts.set(text, content);
+
+      componentsSet.add(text as any);
 
       return text;
     },
@@ -241,14 +252,8 @@ export function createRemoteRoot<
     const normalizedChild =
       typeof child === 'string' ? remoteRoot.createText(child) : child;
 
-    if (!validateAllowedComponent(normalizedChild, components)) {
-      channel(
-        ACTION_ERROR,
-        undefined,
-        ERROR_UNSUPPORTED_COMPONENT,
-        getComponentType(normalizedChild),
-      );
-      return;
+    if (!componentsSet.has(normalizedChild)) {
+      throw new Error('Append invalid component');
     }
 
     return perform(container, {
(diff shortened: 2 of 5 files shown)


packages/core/src/root.ts
```diff
@@ -360,6 +382,19 @@ export function createRemoteRoot<
...
+    component: CanBeChild,
+    allowedComponents: readonly AllowedComponents[],
+  ) {
+    const type = getComponentType(component);
+    const isAllowed = allowedComponents.indexOf(type as any) >= 0;
+    return isAllowed;
+  }
+
+  function getComponentType(component: CanBeChild) {
+    return component.kind === KIND_TEXT ? 'Text' : component.type;
```
lemonmade, line 396: This is not safe because we have components with type `Text` also. IMO, true "text" nodes should always be allowed.

packages/core/src/root.ts
```diff
@@ -31,7 +33,7 @@ export function createRemoteRoot<
   AllowedChildrenTypes extends AllowedComponents | boolean = true
 >(
   channel: RemoteChannel,
-  {strict = true}: Options<AllowedComponents> = {},
+  {strict = true, components = []}: Options<AllowedComponents> = {},
```
lemonmade, line 36: This is technically a breaking change because not providing any components used to mean "all components allowed", but now means "no components allowed". I would prefer if it continued to mean "all components allowed".

packages/core/src/root.ts
```diff
@@ -239,6 +241,16 @@ export function createRemoteRoot<
     const normalizedChild =
       typeof child === 'string' ? remoteRoot.createText(child) : child;
 
+    if (!validateAllowedComponent(normalizedChild, components)) {
```
lemonmade, line 244: I had not actually considered validating when inserting children. It's probably a good idea to validate then, but I think it's more important to validate in `createComponent`. It gives earlier feedback and, if we were to track components created by a given `Root`, would let us safely avoid validating on insertions.

packages/core/src/receiver.ts
```diff
@@ -116,6 +120,9 @@ export class RemoteReceiver {
       text.text = newText;
       this.enqueueUpdate(text);
     },
+    error: () => {
```
lemonmade, line 123: I am not sure what this method is for?

packages/core/src/root.ts
henrytao-me: Agree. I will update it.

packages/core/src/root.ts
henrytao-me: I was looking at `createComponent` before deciding to use append and insert. The problem with `createComponent` is that this function has to return something. In case of unsupported components, what should we do? Should we throw exception to avoid the return value? If so, the whole app won't work anymore. 

With the approach in this PR, the app still works (at least no crash), other supported components still run. The host has option to handle unsupported components in the UI to let partner / merchant know. 

We can discuss tomorrow on a call.

henrytao-me:
Regarding creating new `onError` VS adding new error action with existing pattern, we discussed here https://github.com/Shopify/app-extension-libs/issues/802#issuecomment-675677903. We could chat about it more to see which one makes more sense.

packages/core/src/root.ts
alanthai: Is it possible to return a stub component or just log an error and move on?

vividviolet: Can we not return null?

henrytao-me: The `perform` call below returns void. So, an early return just works.

packages/core/src/root.ts
```diff
@@ -127,9 +132,15 @@ export function createRemoteRoot<
         children.set(component, strict ? Object.freeze([]) : []);
       }
 
+      componentsSet.add(component);
+
       return (component as unknown) as RemoteComponent<typeof type, Root>;
     },
     createText(content = '') {
+      if (!validateAllowedComponent('Text' as any, components)) {
```
lemonmade, line 140: This still gets messy if you have a component called `Text`, which we do at least in checkout's Argo API. I don't think we should ever restrict text nodes.

packages/core/src/root.ts
```diff
@@ -239,6 +252,10 @@ export function createRemoteRoot<
     const normalizedChild =
       typeof child === 'string' ? remoteRoot.createText(child) : child;
 
+    if (!componentsSet.has(normalizedChild)) {
+      throw new Error('Append invalid component');
```
lemonmade, line 256: I think I'd use a more instructive message here, something like `Can’t append a node that was not created by this RemoteRoot`

packages/core/src/root.ts
```diff
@@ -273,6 +290,7 @@ export function createRemoteRoot<
   // Might need to send the removed child ID, or find out if we
   // can collect removals into a single update.
   function removeChild(container: HasChildren, child: CanBeChild) {
+    componentsSet.delete(child);
```
lemonmade, line 293: You shouldn't remove it, there isn't technically anything wrong in removing a child, then re-appending that child to the tree somewhere else

packages/core/src/root.ts
```diff
@@ -360,6 +382,16 @@ export function createRemoteRoot<
           children: value.children.map((child) => serialize(child as any)),
         };
   }
+
+  function validateAllowedComponent(
+    componentType: AllowedComponents,
+    allowedComponents: readonly AllowedComponents[],
+  ) {
+    const isAllowed =
+      allowedComponents.length === 0 ||
```
lemonmade, line 391: I know this is a little pedantic, but this technically prevents passing an empty array to signal "no components are allowed". If you take my suggestion to always allow text nodes, there might be a use case for differentiating "no `components` option passed" (=== allow all components) and "empty `components` option passed" (=== only text nodes are allowed).

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,47 @@
+import {createRemoteRoot} from '../root';
+
+describe('root', () => {
+  it('does not throw error when no allowed components are set', () => {
```
lemonmade, line 4: I'd prefer if these tests were scoped under a `describe('createComponent')`, and we had additional tests under nested `describe('appendChild()')` and `describe('insertChildAfter')` blocks to check the "appending a node not created by this root" validation.

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const components = ['Card'];
+    const root = createRemoteRoot(() => {}, {components});
+
+    expect(() => {
+      const Card = root.createComponent('Card');
+      root.appendChild(Card);
+
+      const Button = root.createComponent('Button');
+      root.appendChild(Button);
+    }).toThrowError();
```
lemonmade, line 43: I think you should only test one component in each assertion, and only test the method you are asserting on (`createComponent`, in all of these). It doesn't matter for the previous two tests, but in this one it's not clear if the test fails "properly", because the error might get thrown from a method that you do not expect.

packages/core/src/root.ts
henrytao-me: Fixed!

packages/core/src/root.ts
henrytao-me: Fixed!

packages/core/src/root.ts
henrytao-me: Fixed!

packages/core/src/tests/root.test.ts
henrytao-me: Fixed!

packages/core/src/tests/root.test.ts
henrytao-me: Fixed!

packages/core/src/root.ts
```diff
@@ -360,6 +381,15 @@ export function createRemoteRoot<
           children: value.children.map((child) => serialize(child as any)),
         };
   }
+
+  function validateAllowedComponent(
+    componentType: AllowedComponents,
+    allowedComponents?: readonly AllowedComponents[],
```
lemonmade, line 387: I think you could technically omit this argument since it's available in scope (unless I am reading incorrectly)

packages/core/src/root.ts
```diff
@@ -147,6 +154,8 @@ export function createRemoteRoot<
       makeRemote(text, id, remoteRoot);
       texts.set(text, content);
 
+      componentsSet.add(text as any);
```
lemonmade, line 157: Why is the cast needed here? Also, given that this contains both texts and components, might be better to just call the set `nodes` or something like that.

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,90 @@
+import {createRemoteRoot} from '../root';
+
+describe('root', () => {
+  describe('createComponent', () => {
```
lemonmade, line 4: Might be nice to have a test that ensures that `createText` is always allowed, even if `components` is defined.

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      expect(() => {
+        const card = {} as any;
+        root.appendChild(card);
+      }).toThrowError();
+    });
+  });
+
+  describe('insertChildBefore', () => {
+    it('does not throw error for component created by insertChildBefore', () => {
```
lemonmade, line 64: `created by the root`?

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      expect(() => {
+        const card = root.createComponent('Card');
+        const button = root.createComponent('Button');
+        const text = root.createComponent('Text');
+        const image = root.createComponent('Image');
+        card.appendChild(button);
+        card.insertChildBefore(text, button);
+        root.appendChild(image);
+        root.insertChildBefore(card, image);
```
lemonmade, line 75: I think you only need one set of `appendChild` + `insertChildBefore` to prove that it works?

henrytao-me:
PR ready for another review 🙇

packages/core/src/root.ts
```diff
@@ -239,6 +248,12 @@ export function createRemoteRoot<
     const normalizedChild =
       typeof child === 'string' ? remoteRoot.createText(child) : child;
 
+    if (!componentsSet.has(normalizedChild)) {
+      throw new Error(
+        `Can't append a node that was not created by this RemoteRoot`,
```
vividviolet, line 253: nit:
> Cannot append a node that was not created by this Remote Root

packages/core/src/root.ts
vividviolet: I think this function is only called once. We might as well just inline the logic above:

```
if (allowedComponents && allowedComponents.indexOf(componentType) < 0) {
  throw new Error(`Unsupported component: ${type}`);
}
```

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      const root = createRemoteRoot(() => {}, {components});
+
+      expect(() => {
+        root.createComponent('Button');
+      }).toThrowError();
+    });
+  });
+
+  describe('appendChild', () => {
+    it('does not throw error for component created by appendChild', () => {
```
vividviolet, line 42: nit: 
> does not throw error when appending a child created by the remote root

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      expect(() => {
+        const card = root.createComponent('Card');
+        const button = root.createComponent('Button');
+        card.appendChild(button);
+        root.appendChild(card);
+      }).not.toThrowError();
+    });
+
+    it('throws error for component which is not created by appendChild', () => {
```
vividviolet, line 53: nit: 
> throws error when appending a child not created by the remote root

packages/core/src/tests/root.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        const text = root.createComponent('Text');
+        const image = root.createComponent('Image');
+        card.appendChild(button);
+        card.insertChildBefore(text, button);
+        root.appendChild(image);
+        root.insertChildBefore(card, image);
+      }).not.toThrowError();
+    });
+
+    it('throws error for component which is not created by insertChildBefore', () => {
```
vividviolet, line 79: Same here,

> throws an error when calling `insertChildBefore` for a component not created by the remote root

PR #25, opened 2020-10-02
introduce remote origin iframe messenger

robin-drexler:
ensure null origin still works as before and remote messenger can be imported without adding the null origin's iframe content to the bundle.


robin-drexler:
Ok, I had to rename iframe to i-frame for now. It seems like github is caching some older build output which still had the `iframe` as a file and not a folder(?) 

Or I'm missing something here.

packages/web-workers/src/messenger/i-frame/create-messenger.ts
```diff
@@ -1,27 +1,14 @@
...
-  </html>
-`;
-
-export function createIframeWorkerMessenger(url: URL): MessageEndpoint {
+export function createIframeWorkerMessenger(
+  url: URL,
+  prepareIframe: (
+    iframe: HTMLIFrameElement,
+    identifier: string,
+  ) => HTMLIFrameElement,
```
lemonmade, line 10: I think I'd drop the requirement to return the `iframe` here

packages/web-workers/src/messenger/i-frame/create-messenger.ts
```diff
@@ -1,27 +1,14 @@
...
-    </body>
-  </html>
-`;
-
-export function createIframeWorkerMessenger(url: URL): MessageEndpoint {
+export function createIframeWorkerMessenger(
+  url: URL,
+  prepareIframe: (
+    iframe: HTMLIFrameElement,
+    identifier: string,
```
lemonmade, line 9: I get why you are passing this through, but since it needs to be manually known for anyone using the non-null-origin iframe method anyways, I think I'd just drop this argument and hardcode it in the null origin file/ export it from a sibling `constants` file instead.

packages/web-workers/src/messenger/i-frame/remote-origin.ts
```diff
@@ -0,0 +1,8 @@
+import {createIframeWorkerMessenger} from './create-messenger';
+
+export function createRemoteOriginIframeMessenger(url: URL, iframeSrc: string) {
```
lemonmade, line 3: I always get a bit worried to see >1 positional argument when all the arguments are primitive (or primitive-like, like `URL`s). I'd probably wrap up that second argument in an options object.

packages/web-workers/src/messenger/index.ts
```diff
@@ -1,2 +1,6 @@
 export {createWorkerMessenger} from './worker';
-export {createIframeWorkerMessenger} from './iframe';
+export {
+  createNullOriginIframeMessenger as createIframeWorkerMessenger,
+  createNullOriginIframeMessenger,
+  createRemoteOriginIframeMessenger,
+} from './i-frame';
```
lemonmade, line 6: `i-frame` is kind of strange...

lemonmade (commented):
I'm good with adding this overall, just some nitpicks 👍

packages/web-workers/src/messenger/index.ts
robin-drexler: [I agree](https://github.com/Shopify/remote-ui/pull/25#issuecomment-702916291). It's self defense for now 😬

lemonmade: I think you can purge the cache by adding a version after the `-typescript-` bit of these lines: https://github.com/Shopify/remote-ui/blob/main/.github/workflows/ci.yml#L39-L41

```
key: ${{ runner.os }}-typescript-v2-${{ github.sha }}
restore-keys: |
  ${{ runner.os }}-typescript-v2-
```

robin-drexler: pushed some updates according to your comments.

I can't seem to change the ci.yml file, though.

```
 ! [remote rejected] iframe-messenger -> iframe-messenger (refusing to allow an OAuth App to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
```

Got to figure out how to fix that first.

robin-drexler: @lemonmade not sure if that's a privilege I may be missing for the repo?

packages/web-workers/src/messenger/index.ts
```diff
@@ -1,2 +1,7 @@
 export {createWorkerMessenger} from './worker';
-export {createIframeWorkerMessenger} from './iframe';
+export {
+  createNullOriginIframeMessenger as createIframeWorkerMessenger,
+  createNullOriginIframeMessenger,
+  createRemoteOriginIframeMessenger,
+  IFRAME_RUN_IDENTIFIER,
```
robin-drexler, line 6: Decided to export the const as folks could use that to generate iFrame host files whenever they update remote-ui.

lemonmade: I don't think you should export this as a public API, so you can safely remove it from here and the `iframe/index.ts` file

lemonmade (approved):

packages/web-workers/src/messenger/iframe/remote-origin.ts
```diff
@@ -0,0 +1,9 @@
+import {createIframeWorkerMessenger} from './create-messenger';
+
+export function createRemoteOriginIframeMessengerFactory(iframeSrc: string) {
```
robin-drexler, line 3: @lemonmade after I used 🎩  in an actual project, I noticed that the signature was off. 

the `createWorkerMessenger` strategy function is actually being called with the URL later by remote-ui itself, but we need to setup the iframeSrc earlier. So I had to  change it to be a factory function.

packages/web-workers/src/messenger/iframe/null-origin.ts
```diff
@@ -0,0 +1,29 @@
...
+
+            var worker = new Worker(workerScript);
+            worker.postMessage({__replace: event.ports[0]}, [event.ports[0]]);
+          });
+        </script>
+      </body>
+    </html>
+  `;
+
+    iframe.setAttribute('sandbox', 'allow-scripts');
```
robin-drexler, line 26: `sandbox allow-scripts` forced remote iframe to be considered a `null` iframe as well, which defeated its purpose, so I moved it to be only applied here.

lemonmade (approved):

PR #28, opened 2020-10-08
pass endpoint options kn createWorker

robin-drexler:
Mainly to allow to pass `callable` to make workers work in IE and other envs that don't support proxies.

### DIFF
--- a/packages/rpc/src/endpoint.ts
+++ b/packages/rpc/src/endpoint.ts
@@ -25,7 +25,7 @@ interface MessageMap {
   [FUNCTION_RESULT]: [string, Error?, any?];
 }
 
-interface Options<T = unknown> {
+export interface CreateEndpointOptions<T = unknown> {
   uuid?(): string;
   createEncoder?(api: EncodingStrategyApi): EncodingStrategy;
   callable?: (keyof T)[];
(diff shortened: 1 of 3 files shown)


packages/rpc/src/endpoint.ts
```diff
@@ -25,7 +25,7 @@ interface MessageMap {
   [FUNCTION_RESULT]: [string, Error?, any?];
 }
 
-interface Options<T = unknown> {
+export interface CreateEndpointOptions<T = unknown> {
```
robin-drexler, line 28: Stayed close to the already existing `CreateWorkerOptions`

packages/web-workers/src/create/worker.ts
```diff
@@ -1,14 +1,20 @@
...
+  MessageEndpoint,
+  CreateEndpointOptions,
+} from '@remote-ui/rpc';
 import {createWorkerMessenger} from '../messenger';
 import {createScriptUrl, FileOrModuleResolver} from './utilities';
 
-export interface CreateWorkerOptions {
+export interface CreateWorkerOptions<T> {
   createMessenger?(url: URL): MessageEndpoint;
+  endpointOptions?: CreateEndpointOptions<T>;
```
robin-drexler, line 12: Not sure if those should be spread into `CreateWorkerOptions ` instead ?

lemonmade: Yeah, I'd rather `CreateWorkerOptions<T> extends CreateEndpointOptions<T>`

lemonmade (approved):

PR #29, opened 2020-10-19
Support type override

henrytao-me:
This changes allows Argo to bypass Polaris type check when rendering StackItem.

Ref https://github.com/Shopify/app-extension-libs/issues/996

### DIFF
--- a/packages/react/src/host/RemoteComponent.tsx
+++ b/packages/react/src/host/RemoteComponent.tsx
@@ -14,7 +14,9 @@ interface Props {
   receiver: RemoteReceiver;
   component: Serialized<RemoteComponentDescription<any, any>>;
   controller: Controller;
-  __type__?: React.ReactChild;
+  // Type override allows components to bypass default wrapping behavior, specifically in Argo Admin which uses Polaris to render on the host. Ex: Stack, ResourceList...
+  // See https://github.com/Shopify/app-extension-libs/issues/996#issuecomment-710437088
+  __type__?: React.ComponentType;
 }
 
 export const RemoteComponent = memo(
--- a/packages/react/src/host/RemoteComponent.tsx
+++ b/packages/react/src/host/RemoteComponent.tsx
@@ -14,6 +14,7 @@ interface Props {
   receiver: RemoteReceiver;
   component: Serialized<RemoteComponentDescription<any, any>>;
   controller: Controller;
+  __type__?: React.ReactChild;
 }
 
 export const RemoteComponent = memo(
@@ -46,6 +47,7 @@ export const RemoteComponent = memo(
               receiver={receiver}
               component={child}
               controller={controller}
+              __type__={(controller.get(child.type) as any)?.__type__}
             />
           );
         } else {
--- a/packages/react/src/host/Renderer.tsx
+++ b/packages/react/src/host/Renderer.tsx
@@ -26,6 +26,7 @@ export const RemoteRenderer = memo(({components, receiver}: Props) => {
           component={child}
           receiver={receiver}
           controller={controller}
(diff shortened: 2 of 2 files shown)


alanthai (approved):
🎩 and working

packages/react/src/host/Renderer.tsx
```diff
@@ -26,6 +26,7 @@ export const RemoteRenderer = memo(({components, receiver}: Props) => {
           component={child}
           receiver={receiver}
           controller={controller}
+          __type__={(controller.get(child.type) as any)?.__type__}
```
vividviolet, line 29: nit: can we make this a variable for readability?

packages/react/src/host/RemoteComponent.tsx
```diff
@@ -46,6 +47,7 @@ export const RemoteComponent = memo(
               receiver={receiver}
               component={child}
               controller={controller}
+              __type__={(controller.get(child.type) as any)?.__type__}
```
vividviolet, line 50: same here, can we add a test?

packages/react/src/host/Renderer.tsx
vividviolet: Can we add a test?

vividviolet (commented):
Nice and simple!

packages/react/src/host/RemoteComponent.tsx
BPScott, line 50: `__type__` is somewhat vague. The code in polaris-react isn't clear that this is due to argo-related functionality. I worry that in 3 month when people come to look at the code in polaris-react they won't have a clue what the extra logic is for and thus might be deleted.

Could this be named more explicitly? A longer name more explicit name would also reduce the potential for accidental naming clashes and thus accidental triggering of the logic in polaris-react. `__remoteUiChildType__` perhaps?

BPScott (commented):
I can't test the logic, but I'd like to see a more explict name so that on the polaris-react side it is more obvious what the code is relating to

packages/react/src/host/Renderer.tsx
henrytao-me: There is no existing test for RemoteComponent. I prefer to skip it for now.

packages/react/src/host/RemoteComponent.tsx
henrytao-me: Same as https://github.com/Shopify/polaris-react/pull/3517#discussion_r508042115

packages/react/src/host/RemoteComponent.tsx
```diff
@@ -14,6 +14,7 @@ interface Props {
   receiver: RemoteReceiver;
   component: Serialized<RemoteComponentDescription<any, any>>;
   controller: Controller;
+  __type__?: React.ReactChild;
```
lemonmade, line 17: This is actually a `ReactComponentType` I think?

lemonmade: I'd also love a comment here that explains why it exists

lemonmade (approved):

packages/react/src/host/RemoteComponent.tsx
henrytao-me: Good catch. Added!

BPScott: With this fix do we need still need the casting in the other files?

henrytao-me: Yes, it does unless we fix the type of the controller too `packages/react/src/host/controller.tsx` but I imagine it would cause more changes than help.

PR #31, opened 2020-10-21
Hot-swap function props

lemonmade:
This is an alternative approach for addressing issues around "stale" function props causing timing issues in some circumstances. I've written a big comment explaining the approach, but basically, instead of relying on the main thread always pointing at the most up-to-date version of a function, we replace every function prop with a wrapper that can we can synchronously update (in the remote context) to point at the most recent version of that prop.

### DIFF
--- a/packages/react/src/reconciler.ts
+++ b/packages/react/src/reconciler.ts
@@ -109,10 +109,7 @@ const reconciler = reactReconciler<
           oldValue
         ) {
           needsUpdate = true;
-          hotSwapFunctions.push([
-            oldValueOnRemoteProps[FUNCTION_CURRENT_IMPLEMENTATION_KEY],
-            newValue,
-          ]);
+          hotSwapFunctions.push([oldValueOnRemoteProps, newValue]);
         }
       }
 
--- a/packages/react/src/reconciler.ts
+++ b/packages/react/src/reconciler.ts
@@ -277,9 +277,16 @@ export default reconciler;
 function wrapFunctionToHotSwapImplementation<T extends Function>(
   func: T,
 ): HotSwappableFunction<T> {
-  const proxyFunction = (...args: any[]) => {
-    return proxyFunction.current(...args);
-  };
-  proxyFunction.current = func;
-  return proxyFunction as any;
+  const wrappedFunction: HotSwappableFunction<T> = ((...args: any[]) => {
+    return wrappedFunction[FUNCTION_CURRENT_IMPLEMENTATION_KEY](...args);
+  }) as any;
+
+  Object.defineProperty(wrappedFunction, FUNCTION_CURRENT_IMPLEMENTATION_KEY, {
+    enumerable: false,
+    configurable: false,
+    writable: true,
+    value: func,
+  });
+
+  return wrappedFunction;
 }
--- a/packages/react/src/reconciler.ts
+++ b/packages/react/src/reconciler.ts
@@ -7,6 +7,12 @@ import type {
   RemoteComponentType,
 } from '@remote-ui/core';
 
+const FUNCTION_CURRENT_IMPLEMENTATION_KEY = '__current';
+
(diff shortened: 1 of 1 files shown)


packages/react/src/reconciler.ts
```diff
@@ -51,20 +57,33 @@ const reconciler = reactReconciler<
...
+  createInstance(type, props, root) {
+    const finalProps: {[key: string]: any} = {};
+
+    for (const key in props) {
+      if (key === 'children') continue;
+
+      const value = props[key];
+
+      finalProps[key] =
+        typeof value === 'function'
```
vividviolet, line 72: Can we put this in a helper like `returnWrappedValue` or something?

packages/react/src/reconciler.ts
```diff
@@ -7,6 +7,12 @@ import type {
   RemoteComponentType,
 } from '@remote-ui/core';
 
+const FUNCTION_CURRENT_IMPLEMENTATION_KEY = '__current';
+
+type HotSwappableFunction<T extends Function> = T & {
+  [FUNCTION_CURRENT_IMPLEMENTATION_KEY]: any;
```
vividviolet, line 13: Can we be more strict than `any` since we know it's 
```
{
    enumerable: false,
    configurable: false,
    writable: true,
    value: T,
}
```

henrytao-me (commented):
🎩-ed locally with app-extension-libs. It doesn't work. This is what I tried

```ts
import React, {useState} from 'react';
import {Button, Stack, TextField} from '@shopify/argo-admin-react';

export function TextFieldExample() {
  const [firstName, setFirstName] = useState('');
  return (
    <Stack>
      <TextField label="firstName" value={firstName} onChange={setFirstName} />
      <Button title="Submit" onPress={() => console.log('submitting', firstName)} />
    </Stack>
  );
}

```

packages/react/src/reconciler.ts
```diff
@@ -73,23 +92,39 @@ const reconciler = reactReconciler<
...
+      // for details on why it is necessary to hot-swap functions).
+      if (typeof oldValue === 'function') {
+        const oldValueOnRemoteProps = (instance.props as any)[key];
+
+        if (
+          oldValueOnRemoteProps?.[FUNCTION_CURRENT_IMPLEMENTATION_KEY] ===
+          oldValue
+        ) {
+          needsUpdate = true;
+          hotSwapFunctions.push([
```
vividviolet, line 112: Do we have to clean this array up at any point? Also, if the extension is passing in a new function every time would we have a performance issue when looping through the array? Ex.
 ```
<Button onPress={() =>{}} />
```

vividviolet (commented):
Looks very promising. I haven't had a chance to test yet - I added some questions

packages/react/src/reconciler.ts
lemonmade: That's the property descriptor; it means that it does definitely start out as `{[FUNCTION_CURRENT_IMPLEMENTATION_KEY]: T}`, but it does not necessarily stay that way. After this is sent over the wire, the original function might change to an `undefined`, in which case `[FUNCTION_CURRENT_IMPLEMENTATION_KEY]: undefined`. That's why I widened the type on this property.

packages/react/src/reconciler.ts
lemonmade: No, these are just collected, then used in `commitUpdate` below this line, at which point they are garbage collected

henrytao-me (approved):
🎩-ed with https://github.com/Shopify/app-extension-libs/compare/issue/remote-ui/timing?expand=1. Looks great.

packages/react/src/reconciler.ts
vividviolet: Oh i see it now, the array is created only when preparing to make update

lemonmade:
Closing in favour of #32

PR #32, opened 2020-10-22
Better core library

lemonmade:
This supersedes https://github.com/Shopify/remote-ui/pull/31. It adds the same function-swapping capabilities as that PR did, but does it for `@remote-ui/core` instead of `@remote-ui/react`. This allows any library built on the core of remote-ui to have the same fixes when closing over state in function props. It also extends this logic to work for array and object props that contain nested functions, not just the functions that appear at the "top level" of the props object.

While I was here, I also rewrote a bunch of the core logic to avoid depending so heavily on `WeakMaps`, and instead have the `readonly` fields on remote nodes (`RemoteComponent#children`, `RemoteText#text`, etc) be implemented through closures instead.

### DIFF
--- a/packages/core/src/root.ts
+++ b/packages/core/src/root.ts
@@ -142,6 +142,9 @@ export function createRemoteRoot<
         get props() {
           return internals.externalProps;
         },
+        get remoteProps() {
+          return internals.internalProps;
+        },
         updateProps: (newProps) =>
           updateProps(component, newProps, internals, rootInternals),
         appendChild: (child) =>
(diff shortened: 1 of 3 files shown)


henrytao-me (commented):
🎩-ed. Works with both vanilla and react. It would be great if it works with custom api that passed through worker.render too.

packages/core/src/root.ts
```diff
@@ -35,369 +68,711 @@ export function createRemoteRoot<
...
+          // only one meaning: the actual, resolved children components and text.
+          //
+          // To enforce this, we delete any prop named "children". We don’t take a copy
+          // of the props for performance, so a user calling this function must do so
+          // with an object that can handle being mutated.
+          if (key === 'children') continue;
+
+          normalizedInitialProps[key] = makeValueHotSwappable(
+            initialProps[key],
+          );
```
alanthai, line 113: How come you don't need to recurse over this?

lemonmade: Good catch, I should be 👍

PR #34, opened 2020-10-29
Make it easier to provide children

lemonmade:
While working on some examples using raw remote-ui objects, I found it annoying that I always had to supply initial children as an array, even if there was only one. This PR changes `RemoteRoot#createComponent` to accept children more flexibly; they can either be provided directly, in the 3rd–nth positional argument spots, or provided as an array in the 3rd positional argument.

### DIFF
--- a/packages/core/src/root.ts
+++ b/packages/core/src/root.ts
@@ -92,7 +92,7 @@ export function createRemoteRoot<
         throw new Error(`Unsupported component: ${type}`);
       }
 
-      const [initialProps, initialChildren] = rest;
+      const [initialProps, initialChildren, ...moreChildren] = rest;
 
       const normalizedInitialProps = initialProps ?? {};
       const normalizedInitialChildren: AnyChild[] = [];
@@ -117,8 +117,22 @@ export function createRemoteRoot<
       }
 
       if (initialChildren) {
-        for (const child of initialChildren) {
-          normalizedInitialChildren.push(normalizeChild(child, remoteRoot));
+        if (Array.isArray(initialChildren)) {
+          for (const child of initialChildren) {
+            normalizedInitialChildren.push(normalizeChild(child, remoteRoot));
+          }
+        } else {
+          normalizedInitialChildren.push(
+            normalizeChild(initialChildren, remoteRoot),
+          );
+
+          // The complex tuple type of `rest` makes it so `moreChildren` is
+          // incorrectly inferred as potentially being the props of the component,
+          // lazy casting since we know it will be an array of child elements
+          // (or empty).
+          for (const child of moreChildren as any[]) {
+            normalizedInitialChildren.push(normalizeChild(child, remoteRoot));
+          }
         }
       }
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -80,6 +80,8 @@ type AllowedTextChildren<
   AllowString extends boolean = false
 > = AllowString extends true ? RemoteText<Root> | string : RemoteText<Root>;
 
+type MaybeArray<T> = T | T[];
+
 export interface RemoteRoot<
   AllowedComponents extends RemoteComponentType<
     string,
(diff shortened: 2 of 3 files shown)


packages/core/src/root.ts
```diff
@@ -117,8 +117,22 @@ export function createRemoteRoot<
...
+          normalizedInitialChildren.push(
+            normalizeChild(initialChildren, remoteRoot),
+          );
+
+          // The complex tuple type of `rest` makes it so `moreChildren` is
+          // incorrectly inferred as potentially being the props of the component,
+          // lazy casting since we know it will be an array of child elements
+          // (or empty).
+          for (const child of moreChildren as any[]) {
+            normalizedInitialChildren.push(normalizeChild(child, remoteRoot));
```
robin-drexler, line 134: I think it's possible that `undefined` gets pushed onto `normalizedInitialChildren` when `initialChild`, but no `moreChildren` are passed.

Is this being sanitized later or could it lead to issues?

robin-drexler (approved):
Love this! 

I'm certain @jared-dykstra will appreciate this as well. :D

packages/core/src/root.ts
robin-drexler: True, I got that confused. When you omit `moreChildren` completely it's an empty array. 👍

packages/core/src/types.ts
```diff
@@ -120,24 +122,34 @@ export interface RemoteRoot<
     ...rest: IfAllOptionalKeys<
       PropsForRemoteComponent<Type>,
       [
-        PropsForRemoteComponent<Type>?,
-        Iterable<
+        (PropsForRemoteComponent<Type> | null)?,
+        MaybeArray<
```
vividviolet, line 126: Can we set the type so that people know to use either the old way or the new way?

```ts
root.createComponent('Stack', undefined, [root.createComponent('Text', undefined, 'This will be fun!')])
```

or 

```ts
root.createComponent('Stack', undefined, root.createComponent('Text', undefined, 'This will be fun!'), ...)
```
Maybe something like:

```ts
[
   (PropsForRemoteComponent<Type> | null)?, 
   [AllowedChildren<
       AllowedChildrenTypes,
        RemoteRoot<AllowedComponents, AllowedChildrenTypes>,
        true
     >]?,
] | 
[
  (PropsForRemoteComponent<Type> | null)?, 
   ...AllowedChildren<
       AllowedChildrenTypes,
       RemoteRoot<AllowedComponents, AllowedChildrenTypes>,
       true
    >[]
]
```

vividviolet (commented):
Love the change

packages/core/src/types.ts
lemonmade: I think this type is correct — this part of the tuple just covers the third positional argument, which can either be an `AllowedChildren<...>` or nothing (hence the `?`). The 4th..nth positional arguments are covered by the "rest" part of the tuple definition (starting with `...` below)

vividviolet: I'm not sure I'm just reading it incorrectly but the current typing does not show an error when you do this:

```ts
root.createComponent('Card', {title: 'a'}, ['1', '2'], '3');
```
If we did the following, I think it will make it so that only either an array of children or positional arguments  is correct:

```ts
createComponent<Type extends AllowedComponents>(
    type: Type,
    ...rest: IfAllOptionalKeys<
      PropsForRemoteComponent<Type>,
      | [
          (PropsForRemoteComponent<Type> | null)?,
          AllowedChildren<
            AllowedChildrenTypes,
            RemoteRoot<AllowedComponents, AllowedChildrenTypes>,
            true
          >[]?,
        ]
      | [
          (PropsForRemoteComponent<Type> | null)?,
          ...AllowedChildren<
            AllowedChildrenTypes,
            RemoteRoot<AllowedComponents, AllowedChildrenTypes>,
            true
          >[]
        ]
    >
  )
```

lemonmade: Thanks, I think I see what you mean now. Can you check to see if the new type for this is what you are expecting?

packages/core/src/root.ts
```diff
@@ -117,8 +117,22 @@ export function createRemoteRoot<
       }
 
       if (initialChildren) {
-        for (const child of initialChildren) {
-          normalizedInitialChildren.push(normalizeChild(child, remoteRoot));
+        if (Array.isArray(initialChildren)) {
```
henrytao-me, line 120: With this check, this case will break 

```
    root.createComponent(
      'Button',
      {
        onPress() {
          console.log('Pressed!');
        },
      },
      ['Press me!'],
      'Press me again!'
    ),
  ]),
```

I think it's better to just normalize both `initialChildren` and `moreChildren` by flattening them. Ex:

```
['foo'], 'bar' => ['foo', 'bar']
'foo', 'bar' => ['foo', 'bar']
'foo', ['bar'] => ['foo', 'bar']
'foo', ['bar'], 'xyz' => ['foo', 'bar', 'xyz']
```

lemonmade: I intentionally wanted to not support this, and the types should not prevent it. If they have an array, that's all we consider (mostly to prevent this being a breaking change); if they have an array and anything else, they can use the spread operator to turn the array into positional arguments.

henrytao-me (approved):
👍

vividviolet (approved):
Code looks good - linting working correctly

PR #35, opened 2020-10-29
More flexible htm integration

lemonmade:
This PR totally reworks the `htm` integration for remote-ui. My first stab at this was weird for a few reasons:

1. You could only append the tree created by the `htm` integration to a `RemoteRoot`, not a `RemoteComponent`
1. The tree of objects you got from calling `htm` was not actually a tree of `RemoteComponent`/ `RemoteText`, it was a tree of an intermediate representation that needed the `render()` call to turn them into "real" instances

This PR removes both of these weirdnesses. You now have to create the tagged template function by passing a `RemoteRoot`. This allows the tree created by the tagged template to be real `RemoteComponent` and `RemoteText` objects. I replaced the `render` function with an `append` function that can append the resulting tree to a `RemoteRoot`, or `RemoteComponent`. This makes it much easier to use the `htm` integration to construct just a part of the tree, which makes this well-suited to creating a large, entirely-static subtree while still using the core remote-ui APIs for the more dynamic bits.

### DIFF
--- a/packages/htm/README.md
+++ b/packages/htm/README.md
@@ -37,9 +37,11 @@ This library only helps you create the initial structure of remote components. A
 `htm` supports [many handy syntax features](https://github.com/developit/htm#syntax-like-jsx-but-also-lit). Notably, if you have external definitions of the remote components available in your system, you can interpolate them as the type name within your markup.
 
 ```tsx
+import {createRemoteRoot} from '@remote-ui/core';
 import {createRender, append} from '@remote-ui/htm';
 import {BlockStack, Button, Text} from 'my-remote-components';
 
+const root = createRemoteRoot(() => {});
 const ui = createRender(root);
 const stack = root.createComponent(BlockStack);
 
--- a/packages/htm/README.md
+++ b/packages/htm/README.md
@@ -18,7 +18,7 @@ npm install @remote-ui/htm --save
 
 ## Usage
 
-`@remote-ui/htm` provides two functions. The first, `createRender`, is a factory function that takes a [`@remote-ui/core` `RemoteRoot`](../core#remoteroot), and returns a [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). This tagged template binds the main function from the `htm` library into creating a tree of `@remote-ui/core` `RemoteComponent` and `RemoteText` objects. `append`, is a helper to call `appendChild()` on a `RemoteRoot` or `RemoteComponent` for each node in the tree returned by the tagged template.
+`@remote-ui/htm` provides two functions. The first, `createRender`, is a factory function that takes a [`@remote-ui/core` `RemoteRoot`](../core#remoteroot), and returns a [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). This tagged template binds the main function from the `htm` library into creating a tree of `@remote-ui/core` `RemoteComponent` and `RemoteText` objects. The second function, `append`, is a helper to call `appendChild()` on a `RemoteRoot` or `RemoteComponent` for each node in the tree returned by the tagged template.
 
 ```ts
 import {createRemoteRoot} from '@remote-ui/core';
--- a/packages/htm/README.md
+++ b/packages/htm/README.md
@@ -29,7 +29,7 @@ const ui = createRender(root);
 
 const onPress = () => {};
 
-append(root, ui`<Button primary onPress=${onPress}>Submit</Button>`);
+append(ui`<Button primary onPress=${onPress}>Submit</Button>`, root);
 ```
 
 This library only helps you create the initial structure of remote components. Any subsequent updates to that structure (for example, responding to a button press by changing its content) would be performed with the [`RemoteRoot` and `RemoteComponent` APIs](../core).
@@ -41,14 +41,13 @@ import {createRender, append} from '@remote-ui/htm';
 import {BlockStack, Button, Text} from 'my-remote-components';
 
 const ui = createRender(root);
+const stack = root.createComponent(BlockStack);
 
 append(
-  root,
   ui`
-    <${BlockStack} spacing="tight">
-      <${Button} onPress=${() => console.log('Pressed!')}>Submit<//>
-      <${Text} subdued>You’ll have a chance to review your purchase<//>
-    <//>
+    <${Button} onPress=${() => console.log('Pressed!')}>Submit<//>
+    <${Text} subdued>You’ll have a chance to review your purchase<//>
   `,
+  stack,
 );
 ```
--- a/packages/htm/README.md
+++ b/packages/htm/README.md
@@ -18,18 +18,18 @@ npm install @remote-ui/htm --save
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


alanthai (approved):
Code looks good. unit test passes

packages/htm/README.md
```diff
@@ -18,34 +18,36 @@ npm install @remote-ui/htm --save
...
+const ui = createRender(root);
+const stack = root.createComponent(BlockStack);
+
+append(
+  ui`
+    <${Button} onPress=${() => console.log('Pressed!')}>Submit<//>
+    <${Text} subdued>You’ll have a chance to review your purchase<//>
   `,
-  root,
+  stack,
```
vividviolet, line 51: nit: Should we add these missing lines to make the example complete?
```
const root = createRemoteRoot(() => {});
....

root.appendChild(stack);
```

packages/htm/src/render.test.ts
```diff
@@ -3,17 +3,19 @@ import {
...
   it('converts htm into RemoteRoot operations', () => {
     const root = createRemoteRoot(() => {});
+    const ui = createRender(root);
+
     const onPress = jest.fn();
-    const content = 'Buy now';
+    const content = 'Buy me!';
 
-    render(htm`<${Button} onPress=${onPress}>${content}<//>`, root);
+    append(ui`<${Button} onPress=${onPress}>${content}<//>`, root);
```
vividviolet, line 18: Can we add another test to verify that using `append` on a non-root component works?

vividviolet (commented):
Great changes - I have some minor suggestions

packages/htm/src/render.ts
```diff
@@ -1,65 +1,38 @@
...
 import type {RemoteRoot, RemoteComponent} from '@remote-ui/core';
 
-interface Element {
-  type: string;
-  props: Record<string, any>;
-  children: (string | Element)[];
-}
+export function createRender(root: RemoteRoot<any, any>) {
+  return baseHtm.bind((type, props, ...children) => {
+    const normalizedChildren: (string | RemoteComponent<any, any>)[] = [];
```
jared-dykstra, line 7: Just stylistic, but especially when currying like this, consider using `=>` for both functions and drop the extra `{ }`?  Prettier is good with line breaks to make it readable.

```ts
const createRender = (root: RemoteRoot<any, any>) =>
  baseHtm.bind((type, props, ...children) => {
    const normalizedChildren: (string | RemoteComponent<any, any>)[] = []; 
    // ...
  });

```

packages/htm/src/render.ts
```diff
@@ -1,65 +1,38 @@
...
+export function append(
+  tree: ReturnType<ReturnType<typeof createRender>>,
+  parent: RemoteRoot<any, any> | RemoteComponent<any, any>,
+) {
+  if (Array.isArray(tree)) {
+    for (const child of tree) {
+      parent.appendChild(child);
     }
+  } else {
+    parent.appendChild(tree);
```
jared-dykstra, line 36: This reads so nicely  :)

packages/htm/README.md
```diff
@@ -18,34 +18,36 @@ npm install @remote-ui/htm --save
...
-import {htm, render} from '@remote-ui/htm';
+import {createRender, append} from '@remote-ui/htm';
 
 const root = createRemoteRoot(() => {});
+const ui = createRender(root);
+
 const onPress = () => {};
 
-render(htm`<Button primary onPress=${onPress}>Submit</Button>`, root);
-root.mount();
```
jared-dykstra, line 31: Is `.mount()` no longer needed?

jared-dykstra (approved):

packages/htm/src/render.ts
```diff
@@ -1,65 +1,38 @@
...
+export function createRender(root: RemoteRoot<any, any>) {
+  return baseHtm.bind((type, props, ...children) => {
+    const normalizedChildren: (string | RemoteComponent<any, any>)[] = [];
 
-export const htm = baseHtm.bind<Element>((type, props, ...children) => {
-  if (typeof type !== 'string') {
-    throw new Error(`You can only create components with string names.`);
-  }
+    for (const child of children) {
+      if (child == null || child === false) continue;
```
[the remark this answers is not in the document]
jared-dykstra: Also, is there a reason `reduce()` isn't used here in place of the `for` loop?  It's accumulating into `normalizedChildren`

packages/htm/README.md
```diff
@@ -18,34 +18,36 @@ npm install @remote-ui/htm --save
...
+`@remote-ui/htm` provides two functions. The first, `createRender`, is a factory function that takes a [`@remote-ui/core` `RemoteRoot`](../core#remoteroot), and returns a [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). This tagged template binds the main function from the `htm` library into creating a tree of `@remote-ui/core` `RemoteComponent` and `RemoteText` objects. The second function, `append`, is a helper to call `appendChild()` on a `RemoteRoot` or `RemoteComponent` for each node in the tree returned by the tagged template.
 
 ```ts
 import {createRemoteRoot} from '@remote-ui/core';
-import {htm, render} from '@remote-ui/htm';
+import {createRender, append} from '@remote-ui/htm';
```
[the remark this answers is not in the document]
lemonmade: I agree that the naming is kind of wonky/ a little all over the place. Ultimately, I wanted to stay away from the explicit `htm` naming in this round because, as I was reading it, it made this library sound too related to HTML, when in fact it has no relationship to that at all. `createRender` was the name that felt best to me (`createUI` made it sound like it was actually creating UI, which it isn't — it's just a factory to create a function that can help with creating UI), and `ui` was the tagged template that felt the most right to me (`render` is very confusing since `@remote-ui/react`/`mini-react` both have a `render` function for starting React's rendering loop). I will give it a bit more thought, though.

packages/htm/README.md
```diff
@@ -18,34 +18,36 @@ npm install @remote-ui/htm --save
...
 
 const root = createRemoteRoot(() => {});
+const ui = createRender(root);
```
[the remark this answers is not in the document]
lemonmade: The `createX(root)` value needs to be the name of the tagged template function — you've moved it to the function that appends, but it needs to be on the tagged template so that it can actually create remote-ui objects.

packages/htm/src/render.ts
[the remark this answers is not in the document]
lemonmade: We use `== null` for this feature pretty often, IMO it's only gotten more acceptable in JS to do so since that's basically the same check that `?.` and `??` use. I could have used a reduce here, but I wanted to keep memory consumption down as much as possible here — which means only creating a single array for holding the children — and I find reduce looks awkward when you just mutate the collector array, instead of creating a new array on each call.

packages/htm/src/render.ts
lemonmade: I think this is stylistic, my preference is strongly for function declarations, especially at module scope, since they get hoisted. I don't feel too strongly about this but I don't see a ton of value in trying to make the code more terse (personally, I find nested array function returns to be quite hard to read, though I know this is not universally true for all developers)

packages/htm/README.md
lemonmade: It is still needed, but I recently removed it from a few other READMEs since we changed Argo to implicitly call this method (not everyone will be using Argo, but I want to make the READMEs support that use case best)

PR #36, opened 2020-11-02
Fix hotswap function when update props

henrytao-me:
Ref https://github.com/Shopify/app-extension-libs/issues/1135

### How to test?
- In https://github.com/shopify/app-extension-libs, do
```
dev up
```
- In this repo, do 
```
yarn && yarn build && cp -R ./packages/* ../app-extension-libs/node_modules/@remote-ui
```
- In https://github.com/shopify/app-extension-libs, do
```
dev up
```
- Verify that ResourceList section works as expected https://argo-playground.myshopify.io/components-list


lemonmade (approved):
Can we add a test for this case?

vividviolet (approved):
Tophat'd 👍

PR #38, opened 2020-11-04
Fix hotswap function when updating arrays of different lengths as nested arrays

vividviolet:
Related to https://github.com/Shopify/app-extension-libs/issues/1142

This fix is really just 2 lines:

image

However, I refactored it a bit so that the code is easier to debug in the future. I think it's slightly easier to read now with less nested conditionals and smaller functions for handling objects and arrays.

### Tophat
1. In https://github.com/shopify/app-extension-libs, do
```
dev up
```
2. In this repo, do 
```
yarn && yarn build && cp -R ./packages/* ../app-extension-libs/node_modules/@remote-ui
```
- In https://github.com/shopify/app-extension-libs, do
```
dev up
```
3. Navigate to https://argo-playground.myshopify.io/components-list 
4. Verify that a modal can be opened with multiple secondary actions
image

### DIFF
--- a/packages/core/src/root.ts
+++ b/packages/core/src/root.ts
@@ -466,127 +466,11 @@ function tryHotSwappingValues(
   }
 
   if (Array.isArray(currentValue)) {
-    if (!Array.isArray(newValue)) {
-      return [
-        makeValueHotSwappable(newValue),
-        collectNestedHotSwappableValues(currentValue)?.map(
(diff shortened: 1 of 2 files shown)


henrytao-me (approved):
🎩-ed all components in Argo Playground. LGTM

lemonmade (approved):
Great job 👏

PR #57, opened 2021-01-16
Use chunkFilename as worker filename when available

henrytao-me:
This PR uses chunkFilename as worker filename when available. It's not possible to add unsupported key to Webpack config (ex: workerFilename) because Webpack validates all the keys. So, I think reusing same chunkFilename for worker file makes sense. 

Resolve https://github.com/Shopify/remote-ui/issues/56

## How to test?
- Run `yarn build`
- Copy all packages to Web via `cp -R ./packages/* ../web/node_modules/@remote-ui/`
- In Web, do 
  + In `scripts/builder/build-argo-host/index.ts`, add this to plugins block
```
        new WebWorkerPlugin({
          filename: isProduction ? '[name].[contenthash].worker.js' : undefined,
        }),
```
  + `IS_CLOUD_ENV=true yarn build:argo-host`
  + Verify that the output file name for worker has content hash.

Note that: running `contenthash` in development mode seems overkill. The webpack task just stops without error and no output in the build folder.

### DIFF
--- a/packages/web-workers/src/webpack-parts/loader.ts
+++ b/packages/web-workers/src/webpack-parts/loader.ts
@@ -89,13 +89,11 @@ export function pitch(
     );
   }
 
+  const {filename, chunkFilename} = compiler.options.output!;
+
   const workerOptions = {
-    filename: addWorkerSubExtension(
-      compiler.options.output!.filename as string,
-    ),
-    chunkFilename: addWorkerSubExtension(
-      compiler.options.output!.chunkFilename!,
-    ),
+    filename: addWorkerSubExtension((chunkFilename || filename) as string),
+    chunkFilename: addWorkerSubExtension(chunkFilename as string),
     globalObject: (plugin && plugin.options.globalObject) || 'self',
   };
 
--- a/packages/web-workers/src/webpack-parts/loader.ts
+++ b/packages/web-workers/src/webpack-parts/loader.ts
@@ -89,11 +89,13 @@ export function pitch(
     );
   }
 
-  const {filename, chunkFilename} = compiler.options.output!;
-
   const workerOptions = {
-    filename: addWorkerSubExtension((chunkFilename || filename) as string),
-    chunkFilename: addWorkerSubExtension(chunkFilename as string),
+    filename: addWorkerSubExtension(
+      plugin.options.filename || (compiler.options.output!.filename as string),
+    ),
+    chunkFilename: addWorkerSubExtension(
+      compiler.options.output!.chunkFilename!,
+    ),
     globalObject: (plugin && plugin.options.globalObject) || 'self',
   };
 
@@ -164,7 +166,7 @@ export function pitch(
 function addWorkerSubExtension(file: string) {
   return file.includes('[name]')
     ? file.replace(/\.([a-z]+)$/i, '.worker.$1')
-    : `${file.split('/').slice(0, -1).join('/')}/[name].worker.js`;
+    : file.replace(/\.([a-z]+)$/i, '.[name].worker.$1');
 }
 
 const loader = {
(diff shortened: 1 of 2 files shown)


packages/web-workers/src/webpack-parts/loader.ts
```diff
@@ -89,13 +89,11 @@ export function pitch(
...
+  const {filename, chunkFilename} = compiler.options.output!;
+
   const workerOptions = {
-    filename: addWorkerSubExtension(
-      compiler.options.output!.filename as string,
-    ),
-    chunkFilename: addWorkerSubExtension(
-      compiler.options.output!.chunkFilename!,
-    ),
+    filename: addWorkerSubExtension((chunkFilename || filename) as string),
```
lemonmade, line 95: I don't think this is the fix we want. It works in the specific case you have, but someone who either doesn't have `[name]` in the `chunkFilename`, or who actually wants to use the `filename` as the base for their worker filename, will not be able to with this approach. `chunkFilename` might also have some substitutions that don't work as a `filename`.

After giving it a bit more thought, I think that the best approach would be to keep using `filename`/ `chunkName` as we were before, but for `addWorkerSubExtension`, we should change the logic to this:

```ts
return file.includes('[name]') ? file.replace(/\.([a-z]+)$/i, '.worker.$1')  : file.replace(/\.([a-z]+)$/i, '.[name].worker.$1');
```

lemonmade (changes requested):

packages/web-workers/src/webpack-parts/loader.ts
henrytao-me: This logic just attempts to append `worker` to the worker file name. It's not possible to add other things like contenthash just for worker file but not the main entry which is what we want for Argo Admin. What option do we have if we want that use case?

vividviolet: Could we add a custom `workerFileName` option to this plugin that the user can pass in? It can default to `compiler.options.output!.filename` like before.

henrytao-me: @vividviolet It's not possible to add unsupported key to Webpack config (ex: workerFilename) because Webpack validates all the keys and throw exception.

vividviolet: Can we not pass the login directly to this plugin when initiating it?

lemonmade: If we need a specific filename for the worker created for argo-admin, I agree with Trish that we should have an option on the loader for it. The default should just do what is usually right, which is to include the hash and a marker that the file is a worker (hence the `.worker.` path part).

henrytao-me: That works 👍

henrytao-me:
This PR is ready to 🚢

packages/web-workers/src/webpack-parts/loader.ts
```diff
@@ -91,7 +91,7 @@ export function pitch(
 
   const workerOptions = {
     filename: addWorkerSubExtension(
-      compiler.options.output!.filename as string,
+      plugin.options.filename || (compiler.options.output!.filename as string),
```
lemonmade, line 94: I think if they provide an explicit `filename`, we should not add the `.worker` to it — that part is up to them. Also, we can use `??` instead of `||` here

lemonmade (changes requested):

packages/web-workers/src/webpack-parts/loader.ts
henrytao-me: 👍 Fixed!

henrytao-me:
🎩-ed again. Still look good

lemonmade (approved):
👍 Make sure to add this option to the README/ CHANGELOG before releasing the new version

vividviolet (approved):
Code looks great but I couldn't tophat due to local issues

PR #58, opened 2021-01-19
Release

henrytao-me:


packages/web-workers/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
```
henrytao-me, line 10: Is this a correct place? Should I create new block for current release version and date? cc @lemonmade

lemonmade: Yes, please create a block for the new version and date

packages/web-workers/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@remote-ui/web-workers",
-  "version": "1.2.9",
+  "version": "1.2.10",
```
lemonmade, line 3: There is technically a new feature in here so IMO this should be `1.3.0` instead

henrytao-me: Fixed!

lemonmade (approved):

PR #68, opened 2021-03-10
Fix receiver state not updating on mount

lemonmade:
`RemoteReceiver#state` was always returning `unmounted`, even after receiving the `mount` method. This PR updates it to return `mounted` when appropriate.

### DIFF
--- a/packages/core/src/receiver.ts
+++ b/packages/core/src/receiver.ts
@@ -95,7 +95,7 @@ export function createRemoteReceiver(): RemoteReceiver {
   >();
 
   let timeout: Promise<void> | null = null;
-  const state: RemoteReceiver['state'] = 'unmounted';
+  let state: RemoteReceiver['state'] = 'unmounted';
 
   const root: RemoteReceiverAttachableRoot = {
     id: ROOT_ID,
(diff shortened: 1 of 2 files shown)


packages/core/src/receiver.ts
```diff
@@ -117,6 +117,8 @@ export function createRemoteReceiver(): RemoteReceiver {
       root.version += 1;
       root.children = normalizedChildren;
 
+      state = 'mounted';
```
kumar303, line 120: To mirror the behavior of `emit('mount')` should this happen within the same [`enqueueUpdate()` wrapper](https://github.com/Shopify/remote-ui/blob/e8f728f6c27d09888f3ba03c4d31a0402e454bca/packages/core/src/receiver.ts#L128-L130)?

kumar303 (approved):
Thanks for jumping on a fix! The code looks good, I just had a non-blocking question.

packages/core/src/receiver.ts
lemonmade: Yeah I kind of flipped back and forth on this, I don't think the library is really handling async updates very elegantly. I ended up matching this mutation to when the root's children are updated, which I think is probably more appropriate than having those two things diverge until the enqueued updates are flushed (by contrast, the event being triggered later feels like less of a problem to me). Definitely need to rethink some parts of this at some point but in the end it works out to the same for all our consuming code :P

kumar303: Sounds good

PR #76, opened 2021-04-28
fix: simplify the stateful subscribable logic

alanthai:
Related: https://github.com/Shopify/checkout-web/pull/4868

Co-authored-by: Henry Tao <henry.tao@shopify.com>

Each unsubscribe will clean up their own logic instead of calling a unified destroy method, which users might forget.

### DIFF
--- a/packages/async-subscription/src/stateful.ts
+++ b/packages/async-subscription/src/stateful.ts
@@ -1,10 +1,5 @@
 import {retain, release} from '@remote-ui/rpc';
-import type {
-  Subscriber,
-  RemoteSubscribable,
-  RemoteSubscribeResult,
-  StatefulRemoteSubscribable,
-} from './types';
+import type {RemoteSubscribable, StatefulRemoteSubscribable} from './types';
 
 export function makeStatefulSubscribable<T>(
   subscription: RemoteSubscribable<T>,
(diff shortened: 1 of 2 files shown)


packages/async-subscription/src/stateful.ts
```diff
@@ -14,46 +9,26 @@ export function makeStatefulSubscribable<T>(
...
 
-  let current = subscription.initial;
-  let listening = true;
-
-  const subscribers = new Set<Subscriber<T>>();
-
-  const subscriptionResult = Promise.resolve<RemoteSubscribeResult<T>>(
-    subscription.subscribe(listener),
-  ).then((result) => {
-    listener(result[1]);
```
alanthai, line 25: Just curious here. Does this `listener` ever reach any subscriptions? That is, does the client have time enough to subscribe before this promise resolves?

packages/async-subscription/src/stateful.ts
```diff
@@ -14,46 +9,26 @@ export function makeStatefulSubscribable<T>(
...
-
-      return () => {
-        subscribers.delete(subscriber);
-      };
+      const subscriptionResult = Promise.resolve(
+        subscription.subscribe((value) => {
+          subscriber(value);
+        }),
+      ).then((result) => {
+        subscriber(result[1]);
```
alanthai, line 24: I think this refactor will cause some tests in checkout-web to fail because the first subscription will automatically call the callback, unlike the previous implementation

PR #86, opened 2021-05-13
Add the ability for consumers to render custom components

james-woo:
Allow consumers of `RemoteRenderer` to pass in their own custom component renderer or text renderer. 

This will allow React Native to override the default behaviour of text rendering, since React Native requires that all Text objects to be wrapped in a ReactElement. 

For web, we can remove the hack that gets the `__type__` out of the component and move it back to web's side by overriding the component rendering.

### DIFF
--- a/packages/react/src/host/RemoteComponent.tsx
+++ b/packages/react/src/host/RemoteComponent.tsx
@@ -1,4 +1,4 @@
-import {memo, useMemo} from 'react';
+import {memo, useMemo, Fragment} from 'react';
 import {
   KIND_COMPONENT,
   KIND_TEXT,
@@ -87,18 +87,24 @@ function renderChildren(
   return [...children].map((child) => {
     switch (child.kind) {
       case KIND_COMPONENT:
-        return renderComponent({
-          component: child,
-          receiver,
-          controller,
-          key: child.id,
-        });
+        return (
+          <Fragment key={child.id}>
+            {renderComponent({
+              component: child,
+              receiver,
+              controller,
+            })}
+          </Fragment>
+        );
       case KIND_TEXT:
-        return renderText({
-          text: child,
-          receiver,
-          key: child.id,
-        });
+        return (
+          <Fragment key={child.id}>
+            {renderText({
+              text: child,
+              receiver,
+            })}
+          </Fragment>
+        );
       default:
         return null;
     }
--- a/packages/react/src/host/RemoteComponent.tsx
+++ b/packages/react/src/host/RemoteComponent.tsx
@@ -1,4 +1,4 @@
-import {memo, useMemo, Fragment} from 'react';
+import {memo, useMemo} from 'react';
 import {
   KIND_COMPONENT,
   KIND_TEXT,
@@ -22,12 +22,14 @@ export function renderComponent({
   component,
   controller,
   receiver,
+  key,
 }: RemoteComponentProps) {
   return (
     <RemoteComponent
       receiver={receiver}
       component={component}
       controller={controller}
+      key={key}
     />
   );
 }
@@ -87,24 +89,18 @@ function renderChildren(
   return [...children].map((child) => {
     switch (child.kind) {
       case KIND_COMPONENT:
-        return (
-          <Fragment key={child.id}>
-            {renderComponent({
-              component: child,
-              receiver,
-              controller,
-            })}
-          </Fragment>
-        );
+        return renderComponent({
+          component: child,
+          receiver,
+          controller,
+          key: child.id,
+        });
       case KIND_TEXT:
-        return (
-          <Fragment key={child.id}>
-            {renderText({
-              text: child,
-              receiver,
-            })}
-          </Fragment>
-        );
+        return renderText({
+          text: child,
+          receiver,
+          key: child.id,
+        });
       default:
         return null;
     }
--- a/packages/react/src/host/RemoteComponent.tsx
+++ b/packages/react/src/host/RemoteComponent.tsx
@@ -1,5 +1,4 @@
 import {memo, useMemo} from 'react';
-import type {ComponentType} from 'react';
 import {
   KIND_COMPONENT,
   KIND_TEXT,
@@ -7,14 +6,12 @@ import {
 } from '@remote-ui/core';
 import type {
   RemoteReceiver,
-  RemoteReceiverAttachableComponent,
   RemoteReceiverAttachableFragment,
   RemoteReceiverAttachableChild,
 } from '@remote-ui/core';
 
-import type {Controller} from './controller';
-import {RemoteText} from './RemoteText';
 import {useAttached} from './hooks';
+import type {Controller, RemoteComponentProps} from './types';
 
 interface RemoteFragmentProps {
   receiver: RemoteReceiver;
@@ -22,20 +19,24 @@ interface RemoteFragmentProps {
   controller: Controller;
 }
 
-interface Props {
-  receiver: RemoteReceiver;
-  component: RemoteReceiverAttachableComponent;
-  controller: Controller;
-  // Type override allows components to bypass default wrapping behavior, specifically in Argo Admin which uses Polaris to render on the host. Ex: Stack, ResourceList...
-  // See https://github.com/Shopify/app-extension-libs/issues/996#issuecomment-710437088
-  __type__?: ComponentType;
-}
-
 const emptyObject = {};
+
+export function renderComponent({
+  component,
+  controller,
+  receiver,
+}: RemoteComponentProps) {
+  return (
+    <RemoteComponent
+      receiver={receiver}
+      component={component}
+      controller={controller}
+    />
+  );
+}
 
 export const RemoteComponent = memo(
   ({receiver, component, controller}: RemoteComponentProps) => {
-    const {renderComponent, renderText} = useRemoteRenderer();
     const Implementation = controller.get(component.type)!;
 
     const attached = useAttached(receiver, component);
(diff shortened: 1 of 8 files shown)


packages/react/src/host/RemoteComponent.tsx
```diff
@@ -41,21 +41,23 @@ export const RemoteComponent = memo(
...
+              });
+              break;
+            case KIND_TEXT:
+              element = renderText({text: child, receiver});
+              break;
+            default:
+              element = null;
+              break;
           }
+          return element ? cloneElement(element, {key: child.id}) : null;
```
lemonmade, line 60: Do we have to clone the element, or could the `renderText`/ `renderComponent` call be expected to do that? This component is rendered for every component in the tree so keeping it as lightweight as possible is definitely desirable

packages/react/src/host/RemoteRenderer.tsx
```diff
@@ -1,41 +1,55 @@
...
-  const {children} = useAttached(receiver, receiver.attached.root)!;
+export const RemoteRenderer = memo(
+  ({
+    controller,
+    receiver,
+    renderComponent = defaultRenderComponent,
+    renderText = defaultRenderText,
+  }: RemoteRendererProps) => {
+    const {children} = useAttached(receiver, receiver.attached.root)!;
+    const renderContextValue = useMemo(() => ({renderText, renderComponent}), [
```
lemonmade, line 26: Does this need to be a separate concept from the `Controller`, which already has a very related job of mapping remote component names => host component implementations?

packages/react/src/host/RemoteRenderer.tsx
```diff
@@ -1,41 +1,55 @@
...
+              });
+              break;
+            case KIND_TEXT:
+              element = renderText({text: child, receiver});
+              break;
+            default:
+              element = null;
+              break;
+          }
+          return element ? cloneElement(element, {key: child.id}) : null;
```
lemonmade, line 50: Same question regarding the need to clone the element

packages/react/src/host/RemoteRenderer.tsx
```diff
@@ -1,41 +1,55 @@
-import {memo} from 'react';
+import {cloneElement, memo, ReactElement, useMemo} from 'react';
```
lemonmade, line 1: `ReactElement` is a type, so should be a type-only import

packages/react/src/host/RemoteRenderer.tsx
```diff
@@ -1,41 +1,55 @@
...
+import {renderComponent as defaultRenderComponent} from './RemoteComponent';
+import {RemoteRendererContext} from './context';
+import type {RemoteComponentProps, RemoteTextProps} from './types';
 
-interface Props {
+export interface RemoteRendererProps {
   receiver: RemoteReceiver;
   controller: Controller;
+  renderComponent?: (props: RemoteComponentProps) => ReactElement;
+  renderText?: (props: RemoteTextProps) => ReactElement;
```
lemonmade, line 15: We usually use method syntax instead of property syntax for these

lemonmade (changes requested):

packages/react/src/host/RemoteRenderer.tsx
henrytao-me: Controller is meant for rendering component type. We also has Text Component as component type which will be rendered using text type. This renderText is for text type. 

Are you saying we should reserve a special Component in components list to render text type? This is what @vividviolet suggested before.

packages/react/src/host/RemoteRenderer.tsx
henrytao-me: It's about the key

packages/react/src/host/RemoteComponent.tsx
lemonmade: Can't `renderComponent` add the key?

packages/react/src/host/RemoteRenderer.tsx
lemonmade: Controller can be whatever we like it to be. Right now, it just gives the component implementation for a remote component, but I think it makes sense for `createController` to accept a second argument with the `renderText` stuff you've added here, and have those properties be exposed publicly such that we don't need a second piece of context.

henrytao-me: We could do it 👍

henrytao-me (approved):
Looks great 👍

packages/react/src/host/controller.tsx
```diff
@@ -1,15 +1,17 @@
...
 }
 
-export interface Controller {
-  get(type: string | RemoteComponentType<string, any, any>): ComponentType<any>;
-}
-
-export function createController(components: ComponentMapping): Controller {
+export function createController(
+  components: ComponentMapping,
+  renderer: Partial<Renderer> = {},
```
lemonmade, line 13: I know it's a little annoying from a typing perspective, but IMO it would be nicer if consumers could pass custom `renderComponent`/ `renderText` functions that will get called with the "default" version of those functions. This is a pattern I've seen in a bunch of other libraries where you might want to customize only a subset of operations, but delegate back to the library for the default behavior in other cases.

lemonmade (changes requested):
This looks really good, just a few niggles/ suggestions left from me

packages/react/src/host/controller.tsx
```diff
@@ -23,9 +32,9 @@ export function createController(
       return value;
     },
     renderer: {
-      renderComponent,
-      renderText,
-      ...renderer,
+      renderComponent:
+        componentRenderer?.(defaultRenderComponent) ?? defaultRenderComponent,
+      renderText: textRenderer?.(defaultRenderText) ?? defaultRenderText,
```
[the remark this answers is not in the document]
lemonmade: I was thinking more like this type signature:

```
interface RendererOptions {
  renderComponent?(details: Parameters<Renderer['renderComponent']>[0], renderDefault: () => ReactNode): ReactNode;
}
```

So a consumer could intentionally return `null` in some cases, and so they are always explicitly returning and not just falling through their function as a way of signalling to use the default. If you guys strongly prefer this approach, it works for me too, I think it's just a slightly more complicated signature since it means the consumer is dealing with some extra closures.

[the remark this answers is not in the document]
henrytao-me: Ah. If you look closer to the commit change, you will see we did exactly what you want but in slightly different way. The consumer absolutely can return null for some cases.

packages/react/src/host/controller.tsx
```diff
@@ -1,16 +1,25 @@
...
+import {renderText as defaultRenderText} from './RemoteText';
 
 export interface ComponentMapping {
   [key: string]: ComponentType<any>;
 }
 
+interface RendererFactory {
+  componentRenderer: (
+    defaultRenderer: Renderer['renderComponent'],
+  ) => Renderer['renderComponent'];
```
henrytao-me, line 14: Specifically this ⬆️ 

This is how the consumer can use the api

```ts
createController(components, {
  componentRenderer: (defaultRenderer) => (props) => {
    const {component} = props;
    switch(component.type) {
       case 'Card': 
         // Don't render Card
         return null;
       case 'Stack':
         // return custom Stack component
         break;
       default:
          return defaultRenderer(props);
    }
  }
});
```

lemonmade: Looking at this with fresh eyes, this API feels a little too complex and verbose. I would prefer this:

```ts
createController(components, {
  renderComponent(component, {key, renderDefault}) {
    switch (component.type) {
      case 'Card': return null;
      case 'Stack': return <CustomStackComponent {...component.props} key={key} />;
      default: return renderDefault();
    }
  }
});
```

We get rid of the "arrow function returning arrow function" business and move some of the arguments around so that `component`/ `text` can be the first positional argument, which lets consumers destructure a little more confidently since they never need to pass the arguments they get back into another function. What do you think?

henrytao-me: Do you think we need to make the usage of key transparent? The consumer of this renderComponent should not know about why they need the key. If they forget to use key, it will complain because the part that needs it is in RemoteRenderer. Other than that, the approach looks good to me

henrytao-me: That should be okay. Let's do it @james-woo. Action items: 
- Use `Fragment` for the key
- Change signature to `renderComponent(component, {renderDefault})`. I think we still keep renderDefault in an object for future extensibility.

henrytao-me: 👍

packages/react/src/host/controller.tsx
```diff
@@ -1,16 +1,45 @@
...
+    renderComponent: externalRenderComponent,
+    renderText: externalRenderText,
+  }: Partial<RendererFactory> = {},
+): Controller {
   const registry = new Map(Object.entries(components));
+  const renderComponent: Renderer['renderComponent'] = externalRenderComponent
+    ? (component) => externalRenderComponent(component, renderComponentOptions)
+    : defaultRenderComponent;
+  const renderText: Renderer['renderText'] = externalRenderText
+    ? (component) => externalRenderText(component, renderTextOptions)
```
lemonmade, line 41:
```suggestion
    ? (text) => externalRenderText(text, renderTextOptions)
```

packages/react/src/host/controller.tsx
```diff
@@ -1,16 +1,45 @@
...
 
 export interface ComponentMapping {
   [key: string]: ComponentType<any>;
 }
 
-export interface Controller {
-  get(type: string | RemoteComponentType<string, any, any>): ComponentType<any>;
+interface RendererFactory {
+  renderComponent(
+    props: RemoteComponentProps,
```
lemonmade, line 20: I still really wish the component itself was alone as the first argument of these functions, since you will always need to pull that off this argument, but can't do it by destructuring if you end up needing to call `renderDefault()`...

packages/react/src/host/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  controller: Controller;
+}
+
+export interface Controller {
+  get(type: string | RemoteComponentType<string, any, any>): ComponentType<any>;
+  renderer: Renderer;
+}
+
+export interface Renderer {
+  renderComponent(props: RemoteComponentProps): ReactElement;
```
lemonmade, line 35: I was getting super confused by the use of the word "props" here.

packages/react/src/host/controller.tsx
```diff
@@ -1,16 +1,45 @@
...
+    options: RenderComponentOptions,
+  ): ReactElement;
```
lemonmade, line 22: Can these not also return falsy values/ `ReactNode`s?

lemonmade (approved):

packages/react/src/host/controller.tsx
lemonmade: `renderDefault` could in theory be made to take no arguments, couldn't it?

henrytao-me: Yes, we could. 👍 James is making a change

PR #97, opened 2021-06-03
Adds iframe-related adaptors

lemonmade:
This PR adds a few iframe-related adaptors to `@remote-ui/rpc`. I was working on a "web-native" demo — including no bundling — and for that purpose, workers were a bit annoying, because they do not widely support native ESM. Instead, I opted to use an iframe as the sandbox. That works great, but there is a bunch of annoying boilerplate for wrapping an iframe as an endpoint on both the parent and child. This PR abstracts that boilerplate into a couple of reusable adaptors.


henrytao-me (approved):
LGTM

PR #103, opened 2021-06-18
Rethrow error

henrytao-me:
This PR rethrows error after sending it to the host endpoint. This will help with analytics in WebWorker. See https://github.com/Shopify/checkout-web/issues/5252#issuecomment-863488300


lemonmade (approved):
👍 not sure what you were thinking for the version bump, but IMO this is a bugfix so I'd probably do a patch change.

vividviolet (approved):

PR #104, opened 2021-06-21
Publish

henrytao-me:


packages/rpc/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.2.2] - 2021-06-21
+
+- Fixed error handling in rpc layer ([pull request](https://github.com/Shopify/remote-ui/pull/103)).
```
lemonmade, line 12: I think we should be a bit more specific about what the fix was (also, if we use RPC outside the library name, we should capitalize it, since it's an acronym)

lemonmade (approved):

lemonmade (approved):

PR #105, opened 2021-06-25
Fix worker webpack build

lemonmade:
This PR fixes some missing files from the builds for this library. Before https://github.com/Shopify/remote-ui/pull/91, `sewing-kit` used Babel to transpile all the source files directly into the build directory. After that change, we are using a newer, rollup-based build system. This system has many advantages, but it does not compile code unless that code is part of the dependency graph of the library's entries. `web-workers/src/webpack-parts/loader.ts` was never imported by anything; instead, the Babel plugin for this library assumes the built version of that file will be present, and references it as part of a webpack loader statement (e.g., `import webworker from '/Users/foo/app/node_modules/@remote-ui/web-workers/dist/cjs/webpack-parts/loader.js!./worker'`).

This PR fixes this issue by referencing the loader as part of the `webpack-parts` entry, which forces the file to be compiled. I also removed an excess `webpack-parts.ts` file I found that seemed to be from an older version of the library.

### DIFF
--- a/packages/web-workers/sewing-kit.config.ts
+++ b/packages/web-workers/sewing-kit.config.ts
@@ -1,5 +1,8 @@
 import {createPackage, Runtime} from '@sewing-kit/config';
+import {createProjectBuildPlugin} from '@sewing-kit/plugins';
 import {defaultProjectPlugin} from '../../config/sewing-kit';
+
+import type {} from '@sewing-kit/plugin-rollup';
 
 export default createPackage((pkg) => {
   pkg.runtimes(Runtime.Node, Runtime.Browser);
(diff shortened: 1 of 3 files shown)


henrytao-me (changes requested):
`loader` is fine now. However, `expose.js.raw` is also missing from the output https://unpkg.com/browse/@remote-ui/web-workers@1.4.1/build/cjs/wrappers/

lemonmade:
@henrytao-me ready for another look

henrytao-me (approved):
🎩-ed. LGTM 👍

packages/web-workers/sewing-kit.config.ts
```diff
@@ -1,12 +1,15 @@
 import {createPackage, Runtime} from '@sewing-kit/config';
+import {createProjectBuildPlugin} from '@sewing-kit/plugins';
 import {defaultProjectPlugin} from '../../config/sewing-kit';
 
+import type {} from '@sewing-kit/plugin-rollup';
```
henrytao-me, line 5: What is this one for?

PR #112, opened 2021-07-26
Deduplicate @types/react

henrytao-me:
Follow up of https://github.com/Shopify/remote-ui/pull/111/files#r674172103

## Context: 
`types/react` was moved from peerDependencies to dependencies since this PR https://github.com/Shopify/remote-ui/pull/108/files. However, I don't understand why yarn.lock isn't updated until now 🤔

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3094,16 +3094,7 @@
   dependencies:
     "@types/react" "*"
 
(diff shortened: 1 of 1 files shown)


lemonmade (approved):
Thank you!

PR #134, opened 2022-01-10
Store container reference to allow multiple render calls

clauderic:
This PR stores the React Reconciler container associated to a remote root in a WeakMap.

Calling `render` multiple times is the only way to update the root React element. Without this change, calling `render` multiple times for the same root creates a new container on every invocation.

### DIFF
--- a/packages/react/src/render.tsx
+++ b/packages/react/src/render.tsx
@@ -1,26 +1,51 @@
 import type {ReactElement} from 'react';
+import type {RemoteRoot} from '@remote-ui/core';
+import type {RootTag} from 'react-reconciler';
 
-import reconciler from './reconciler';
+import {reconciler} from './reconciler';
+import type {Reconciler} from './reconciler';
 import {RenderContext} from './context';
+import type {RenderContextDescriptor} from './context';
+
+const cache = new WeakMap<
+  RemoteRoot,
+  {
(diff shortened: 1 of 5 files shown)


clauderic:
@lemonmade I think this should be pretty much good to go, though feel free to take a final look. 

How do you manage releases of @remote-ui? Anything I should add to this PR in that regard?

packages/react/src/render.tsx
```diff
@@ -1,26 +1,49 @@
 import type {ReactElement} from 'react';
+import type {RemoteRoot} from '@remote-ui/core';
+import type {RootTag} from 'react-reconciler';
 
-import reconciler from './reconciler';
-import {RenderContext} from './context';
+import {reconciler, Reconciler} from './reconciler';
```
lemonmade, line 5: I think we want the `Reconciler` to be brought in with `import type` here

lemonmade (approved):
You're good to merge once the `import type` change is in. I can release the new version once it's in. I could see an argument for any version bump here, but I'd probably do a minor bump — I consider the previous behavior of calling `render` multiple times to have been undesigned, so nothing to break :P What do you think?

clauderic:
I think a minor bump is probably fine here, it feels appropriate to label this as a new feature considering this was an unhandled case in the past – I don't think this is breaking any existing functionality.

PR #138, opened 2022-02-05
Fix rendering a single child with fragment props

jt-stripe:
When rendering a single child of a component that takes fragment props,
the wrapper would error when trying to expand the children array when
the array was not an array and was instead a single child.

React provides `React.Children.array` for exactly this purpose, so we
wrap the use of `externalChildren` in that and add a test that would
previously fail and now succeeds.


lemonmade:
This fix has been published in `@remote-ui/react@4.4.2`. Thanks again for the fix, @jt-stripe (and friends)!

jt-stripe:
🙌🎉

PR #139, opened 2022-02-08
Update RPC Readme

t6d:
Fixes the code examples in the adapters section. The code examples failed to highlight that the result returned from the adapters has to subsequently converted into an Endpoint through the use of `createEndpoint()`. This PR updates the adaptors section in the RPC Readme to highlight that fact.


t6d:
@lemonmade, I was initially also considering to update

> This library also provides a collection of adaptors that transform common `postMessage`-related objects into an `Endpoint`. 

to

> This library also provides a collection of adaptors that transform common `postMessage`-related objects into a `MessageEndpoint` and subsequently into an `Endpoint`.

However, I don't think given the current naming of the types that would make understanding what you have to do easier.

Should we rename `MessageEndpoint` to `MessageAdapter` or something similar to indicate that it is a lower level construct than `Endpoint`?

PR #140, opened 2022-02-08
@remote-ui/testing: Fix an issue where find() and findAll() cannot find components with fragment props

ericfrank-stripe:
Components with fragment props are wrapped in a `ComponentWrapper`, so they don't import as strings, which is what the `find()` and `findAll()` test methods are expecting (see [the relevant code](https://github.com/jt-stripe/remote-ui/blob/main/packages/react/src/components.tsx#L24-L27)). This adds a `getTypeString` function that returns the `type.displayName` if the component is wrapped in a `ComponentWrapper`.

I've added tests for this as well—the first set of tests in the `testing` package.

### DIFF
--- a/packages/testing/CHANGELOG.md
+++ b/packages/testing/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed an issue where `find()` and `findAll()` couldn't find components with fragment props.
+
 ## [1.0.3] - 2020-06-24
 
 ### Fixed
--- a/packages/testing/loom.config.ts
+++ b/packages/testing/loom.config.ts
@@ -6,5 +6,5 @@ export default createPackage((pkg) => {
   pkg.runtimes(Runtime.Node);
   pkg.entry({root: './src/index.ts'});
   pkg.entry({name: 'matchers', root: './src/matchers/index.ts'});
-  pkg.use(defaultProjectPlugin());
+  pkg.use(defaultProjectPlugin({react: true}));
 });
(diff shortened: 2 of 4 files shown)


lemonmade (changes requested):
I think this behavior is actually intentional. I didn't want to tie the `@remote-ui/testing` library to React; it was meant to primarily help developers working with the "vanilla" API of `@remote-ui/core`, but should also work for a developer using any of the "remote" libraries (core, react, react-mini, and vue).

If a React consumer wants to test their application, I think there are two existing options that I'd rather push you towards rather than make this change:

You can use `@remote-ui/testing`, but use the `displayName` of the component when you know you are dealing with a function React component:

```tsx
import {mount} from '@remote-ui/testing';
import {render, createRemoteReactComponent} from '@remote-ui/react';

const FragmentPropsComponent = createRemoteReactComponent(/* ... */);

const wrapper = mount((root) => render(<FragmentPropsComponent />, root));
expect(wrapper.find(FragmentPropsComponent.displayName)).not.toBeNull();
```

Alternatively, you can use a React testing library that can mount a React tree using `react-test-renderer`, which works great with any kind of custom React reconciler-based library, like `@remote-ui/react`. On my current project at Shopify, we use [`@quilted/react-testing`](https://github.com/lemonmade/quilt/tree/main/packages/react-testing) for this, which is a custom version of our [main testing library](https://github.com/Shopify/quilt/tree/main/packages/react-testing) I wrote with support for `react-test-renderer`. It has a more direct set of component-focused testing APIs, and was the basis for the `@remote-ui/testing` APIs:

```tsx
import {mount} from '@quilted/react-testing';
import {createRemoteReactComponent} from '@remote-ui/react';

const FragmentPropsComponent = createRemoteReactComponent(/* ... */);

const wrapper = mount(<FragmentPropsComponent />);
expect(wrapper).toContainReactComponent(FragmentPropsComponent);
```

PR #145, opened 2022-02-22
Remove unnecessary useReceiver() and useController() hooks from the React library

lemonmade:
https://github.com/Shopify/remote-ui/issues/142 asked what the purpose of the `useReceiver()` and `useController()` APIs was. The answer was "nothing" — these utilities were used by the library in earlier versions, but are no longer used internally, and have no real value externally. This PR removes them.


vividviolet (approved):
Looks good to me 👍

PR #147, opened 2022-03-08
allow ArrayBuffer to be sent over rpc layer

robin-drexler:
In https://github.com/Shopify/remote-ui/pull/146, @mathieulag and I were looking at being able to use `File`s as function parameters. @lemonmade rightfully pointed out that `File` is a web only api and `@remote-ui/rpc` shouldn't rely on web only APIs.

In this attempt we're using and transferring ArrayBuffers which can be used to construct `File`s and can also be extracted from a `File` object. 

We were also looking into using a `FileLike` object when communicating between the worker and the main page and ways to convert these to and from `File` for usage with 3rd party code that relies on `File`.

```ts

  interface FileLike {
    metadata: {
      name: string;
      lastModified: number;
      type: string;
    };
    content: ArrayBuffer;
  }
```

With this approach, the sending side will always lose ownership of the `ArrayBuffer`, which may be surprising to developers. 

To mitigate this, we could ensure to always provide APIs on the worker side, which ["clone"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice) the buffer before actually passing it. 

Alternatively, we could also not transfer the `ArrayBuffer`, although this would come with a performance hit: https://developers.google.com/web/updates/2011/12/Transferable-Objects-Lightning-Fast

Which may actually be less severe than cloning the `ArrayBuffer` every time.

### DIFF
--- a/packages/rpc/CHANGELOG.md
+++ b/packages/rpc/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Allow `ArrayBuffer` to be sent over the rpc layer ([pull request](https://github.com/Shopify/remote-ui/pull/147)).
+
 ## [1.2.6] - 2022-01-25
 
 - Stopped throwing an error in some `postMessage` event handlers, which prevents unhandled promise rejection listeners from running unnecessarily.
--- a/packages/rpc/src/encoding/basic.ts
+++ b/packages/rpc/src/encoding/basic.ts
@@ -63,6 +63,10 @@ export function createBasicEncoder(api: EncodingStrategyApi): EncodingStrategy {
         return [value];
       }
 
+      if (value instanceof ArrayBuffer) {
+        return [value];
+      }
+
       const transferables: Transferable[] = [];
 
       if (Array.isArray(value)) {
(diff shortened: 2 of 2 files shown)


lemonmade:
Add a changelog entry and open the PR for real and we can ✅ 🚢

lemonmade (approved):
I like the change 👍 What are you thinking as the version bump for `@remote-ui/rpc`? I think you could make the case for any bump type, but to me this feels like it is a new feature the library is adding support for, so a minor fix feels best.

robin-drexler:
>  it is a new feature the library is adding support for, so a minor fix feels best

yeah, sounds good to me. :) 
I guess we could also consider it a bugfix depending on where you stand, but new feature sounds good as well!

Do you need me to do anything or can / will you do the release? 🙏

robin-drexler:
@lemonmade not sure if you're notified of merges automatically. But this PR is in main now. :)

lemonmade:
👍 Released as `@remote-ui/rpc@1.3.0` (which also patch-bumped all the other packages)

robin-drexler:
❤️ Thanks a lot, @lemonmade !

lemonmade:
Thank you for contributing the feature, I love that we are pushing the boundaries of what this library is able to do!

PR #148, opened 2022-03-17
Support React.cloneElement for Host component

henrytao-me:
This PR adds support for React.cloneElement on the host.


packages/react/src/host/RemoteComponent.tsx
```diff
@@ -69,11 +70,11 @@ export const RemoteComponent = memo(function RemoteComponent({
   const {children} = attached;
 
   if (children.length === 0) {
-    return <Implementation {...props} />;
+    return <Implementation {...otherProps} {...props} />;
   }
 
   return (
-    <Implementation {...props}>
+    <Implementation {...otherProps} {...props}>
```
henrytao-me, line 77: We don't want `props` to be overridden. So, I put `otherProps` in front. Also, there is no need to do the same for Fragment and Text.

lemonmade (changes requested):
Discussed why I am not a big fan of this change in https://github.com/Shopify/ui-extensions-private/discussions/1681 (private repo, sorry non-Shopifolk). TL;DR: it is a hack around APIs that try to use "components as configuration", which I feel is a totally broken pattern in React that we should not bring to UI extensions.

PR #150, opened 2022-03-31
Improve renderComponent() utility in React to make "position in tree" restrictions a little easier

lemonmade:
This PR adds a new field to the `renderComponent()` function we introduced in https://github.com/Shopify/remote-ui/pull/86. The new field is `parent`, which gives the host the ability to easily customize behavior based on what other component is rendering this one. This is particularly useful if you want to enforce some components being allowed as direct children of the remote root, and others only being allowed if they are nested below the root.

We never added documentation to `renderComponent()`, so I have done that here.

cc/ @js-goupil this adds what I think is the missing piece for easily implementing more advanced component restrictions on the host, if you are interested in doing that instead of a multi-extension point setup.

### DIFF
--- a/packages/core/src/receiver.ts
+++ b/packages/core/src/receiver.ts
@@ -8,6 +8,7 @@ import {
   ACTION_UPDATE_TEXT,
   KIND_COMPONENT,
   KIND_FRAGMENT,
+  KIND_ROOT,
 } from './types';
 import type {
   ActionArgumentMap,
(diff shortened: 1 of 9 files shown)


henrytao-me (approved):
This is awesome 🙇

lemonmade:
@kumar303 there were only some e2e tests, but I added in a few more for these behaviors 👍

PR #156, opened 2022-05-27
Remove deprecated folder path mapping in `exports` field

marvinhagemeister:
This PR removes the deprecated `"./": "./"` path mappings in the `exports` field inside `package.json`.

Link to official deprecation notice in node: 607d48e7987200fa084d1864a95374ab5cd93373


lemonmade (approved):
❤️ Thank you for this!

PR #158, opened 2022-06-02
Add a failing test to showcase duplicate node bug

cchalstrom-stripe:
# Add failing tests
This adds some failing tests to showcase what seems like a bug in remote-ui to me.

In the first test, we create a remote app that calls the `useState` hook with an array.  We then call `setState` with a new array, reversing the items.  However, the remote tree does not render as expected afterwards.  It seems that calling `setState` with an array does not re-render the component.

The issue raised in the second test may be related.  It is similar to the first test, except it also uses state of a number to reliably re-render the remote tree.  After re-rendering, we find that list items have been duplicated and 5 child nodes are present instead of 3.

### DIFF
--- a/packages/react/src/tests/e2e.test.tsx
+++ b/packages/react/src/tests/e2e.test.tsx
@@ -417,49 +417,6 @@ describe('@remote-ui/react', () => {
   });
 
   describe('Remote tree updates', () => {
-    it('updates the remote tree when setting state with an array', () => {
-      const receiver = createRemoteReceiver();
-      const remoteRoot = createRemoteRoot(receiver.receive, {
-        components: [RemoteBox],
-      });
-
-      function RemoteApp() {
-        const [items, setItems] = useState(['a', 'b', 'c']);
-        const handleClick = useCallback(() => {
-          setItems((prev) => prev.reverse());
-        }, []);
-        return (
-          <RemoteBox id="click-target" onClick={handleClick}>
-            {items.map((item) => (
-              <RemoteBox key={item}>{item}</RemoteBox>
-            ))}
-          </RemoteBox>
-        );
-      }
-
-      const controller = createController({Box: HostBox});
-
-      function HostApp() {
-        return <RemoteRenderer controller={controller} receiver={receiver} />;
-      }
-
-      domAct(() => {
-        domRender(<HostApp />, appElement);
-        render(<RemoteApp />, remoteRoot, () => {
-          remoteRoot.mount();
-        });
-        jest.runAllTimers();
-        Simulate.click(appElement.querySelector('#click-target')!);
-        jest.runAllTimers();
-      });
-
-      const children = appElement.querySelector('#click-target')!.childNodes;
-      expect(children).toHaveLength(3);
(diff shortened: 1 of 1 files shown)


packages/react/src/tests/e2e.test.tsx
```diff
@@ -391,4 +415,94 @@ describe('@remote-ui/react', () => {
...
+    it('updates the remote tree when setting state with an array', () => {
+      const receiver = createRemoteReceiver();
+      const remoteRoot = createRemoteRoot(receiver.receive, {
+        components: [RemoteBox],
+      });
+
+      function RemoteApp() {
+        const [items, setItems] = useState(['a', 'b', 'c']);
+        const handleClick = useCallback(() => {
+          setItems((prev) => prev.reverse());
```
marvinhagemeister, line 429: FYI: This does not create a new array. The [Array.prototype.reverse()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse) method mutates the array in place. This can be verified in the console:

```js
const a = [1, 2, 3];
console.log(a === a.reverse()); // Logs: true
```

This is the reason why no state update is triggered. The setter functions use a referential equality check under the hood (`===`) and due to `.reverse()` mutating the array in place the setter receives the same array as it already had and cancels the state update.

cchalstrom-stripe:
@marvinhagemeister good call, that makes sense.  Thanks for pointing that out.  I still think there is an issue though, when I setState with a brand new array, elements are duplicated:

```
  ● @remote-ui/react › Remote tree updates › renders the expected items after state updates

    expect(received).toHaveLength(expected)

    Expected length: 3
    Received length: 5
    Received object: [<div>a</div>, <div>b</div>, <div>c</div>, <div>b</div>, <div>a</div>]

      455 |
      456 |       const children = appElement.querySelector('#click-target')!.childNodes;
    > 457 |       expect(children).toHaveLength(3);
          |                        ^
      458 |       expect(children[0].textContent).toBe('c');
      459 |       expect(children[1].textContent).toBe('b');
      460 |       expect(children[2].textContent).toBe('a');

      at Object.<anonymous> (src/tests/e2e.test.tsx:457:24)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       1 failed, 21 passed, 22 total
Snapshots:   0 total
Time:        1.943 s, estimated 2 s
Ran all test suites related to changed files in 2 projects.
```

It seems like it could be a problem in the reconciler config to me?  This issue happens when I use `key` with an arbitrary string ID, but not when I use `key` with the array index of the item.

packages/react/src/tests/e2e.test.tsx
cchalstrom-stripe: Thanks!  I deleted a test which was no longer relevant after this fix.  I still think there could be an issue with rendering duplicate nodes on the host side of the tree, though.  When I make this change, it renders:

```
<div>a</div>
<div>b</div>
<div>c</div>
<div>b</div>
<div>a</div>
```

cchalstrom-stripe:
Hooray, thanks!  I'll close this out.

PR #160, opened 2022-06-03
Fix child reordering in `@remote-ui/react`

lemonmade:
Fixes https://github.com/Shopify/remote-ui/issues/159. Before this PR, there was an issue when React attempted to reorder an existing node among its children. This typically happens when you use keys, as their purpose is to teach React which elements are the same "instance" between renders.

This issue was happening because of a bug in the `@remote-ui/core` library; when appending a child to a parent, we were correctly detaching that child from any existing parent locally (that is, in the remote context), but we were not sending the update to remove that child from its current parent to the host. This PR fixes that by sending the appropriate message to the host when this situation is detected.

cc/ @cchalstrom-stripe who brought this up first, and @alxclark who happened to find it at basically the same time on a Shopify project :P

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -34,7 +34,6 @@ export type {
 } from './types';
 export {createRemoteRoot} from './root';
 export {
-  ROOT_ID,
   createRemoteReceiver,
   createRemoteChannel,
   isRemoteFragmentSerialization,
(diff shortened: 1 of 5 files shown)


cchalstrom-stripe:
:shipit: 
Awesome!  Thanks for the fix.  I was hunting in the right place, but I'm not sure I could have fixed that so easily.

marvinhagemeister (approved):

PR #161, opened 2022-06-08
Fix more remote node reordering issues

lemonmade:
My previous fix in https://github.com/Shopify/remote-ui/pull/160 only covered a subset of the reordering bugs in remote-ui. There were two other issues that caused problems in other cases:

1. We were not sending the correct index to `appendChildBefore()` in cases where the child was moving to a larger index in `children` in the same parent as it was already attached
2. We were not replicating the "append starts by removing from an existing node" behavior in the `RemoteRoot` object.

This PR fixes both of those issues. The code is a little gnarly and I feel like I could do the whole "map remote root to remote receiver" much more cleanly if I started from scratch, but it appears to fix the reordering bugs that the last PR did not.

### DIFF
--- a/packages/core/src/receiver.ts
+++ b/packages/core/src/receiver.ts
@@ -146,16 +146,46 @@ export function createRemoteReceiver(): RemoteReceiver {
         emit('mount');
       });
     },
-    insertChild: (id, index, child) => {
+    insertChild: (id, index, child, existingId) => {
       const attached = attachedNodes.get(
         id ?? ROOT_ID,
       ) as RemoteReceiverAttachableRoot;
-
-      const normalizedChild = normalizeNode(child, addVersion);
-      retain(normalizedChild);
-      attach(normalizedChild);
-
       const {children} = attached;
+
+      let existingAttached: RemoteReceiverAttachableRoot | undefined;
+      let normalizedChild: RemoteReceiverAttachableChild;
+
+      if (id === existingId) {
+        existingAttached = attached;
+      } else if (existingId !== false) {
+        existingAttached = attachedNodes.get(
+          existingId ?? ROOT_ID,
+        ) as RemoteReceiverAttachableRoot;
+      }
+
+      if (existingAttached) {
+        const childId = child.id;
+        const existingChildren = existingAttached.children;
+        const existingIndex = existingChildren.findIndex(
+          (child) => child.id === childId,
+        );
+
+        const [removed] = existingChildren.splice(existingIndex, 1);
+
+        normalizedChild = removed;
+
+        // If we are just moving the child to a different index in the same node,
+        // we don’t need to enqueue an update, because that will be done for this
+        // node below.
+        if (id !== existingId) {
+          existingAttached.version += 1;
+          enqueueUpdate(existingAttached);
+        }
+      } else {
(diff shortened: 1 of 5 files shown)


alxclark (approved):
I tophatted the changes using my own internal app and this fixed all of the rendering issues I was noticing previously. Thanks for the fix!

ericfrank-stripe:
@alxclark Could you give a brief description of the issues you were hitting? I'm curious if some of the bugs we're seeing at Stripe are related.

packages/core/src/receiver.ts
```diff
@@ -146,16 +146,46 @@ export function createRemoteReceiver(): RemoteReceiver {
...
+      let existingAttached: RemoteReceiverAttachableRoot | undefined;
+      let normalizedChild: RemoteReceiverAttachableChild;
 
-      const {children} = attached;
+      if (id === existingId) {
+        existingAttached = attached;
+      } else if (existingId !== false) {
+        existingAttached = attachedNodes.get(
+          existingId ?? ROOT_ID,
+        ) as RemoteReceiverAttachableRoot;
```
kumar303, line 163: This case doesn't seem to have test coverage. Is it easy to add?

packages/core/src/receiver.ts
```diff
@@ -146,16 +146,46 @@ export function createRemoteReceiver(): RemoteReceiver {
...
+
+        normalizedChild = removed;
+
+        // If we are just moving the child to a different index in the same node,
+        // we don’t need to enqueue an update, because that will be done for this
+        // node below.
+        if (id !== existingId) {
+          existingAttached.version += 1;
+          enqueueUpdate(existingAttached);
+        }
```
kumar303, line 183: I don't see any test failures when commenting this out. Maybe there is a way to cover it?

packages/core/src/receiver.ts
```diff
@@ -146,16 +146,46 @@ export function createRemoteReceiver(): RemoteReceiver {
...
+      } else {
+        normalizedChild = normalizeNode(child, addVersion);
+        retain(normalizedChild);
+        attach(normalizedChild);
```
kumar303, line 187: I don't see test failures when commenting this out, either

packages/core/src/root.ts
```diff
@@ -623,24 +627,45 @@ function appendChild(
...
+        currentChildren.splice(existingIndex, 1);
+
+        if (currentParent === container) {
+          newChildren = currentChildren;
+        } else {
+          currentInternals.children = strict
+            ? Object.freeze(currentChildren)
+            : currentChildren;
+
+          newChildren = [...internals.children];
```
kumar303, line 661: I don't see test failures when commenting this out

packages/core/src/root.ts
```diff
@@ -699,21 +724,43 @@ function insertChildBefore(
...
```
kumar303, line 758: I don't see test failures when commenting this out

kumar303 (approved):
If it works, ship it 🤷 

It hurts my brain but thanks for adding tests. As you mentioned, it would be nice to simplify this logic later on. Because of that, it would help to have solid test coverage. I commented on some bits that look to be missing test coverage, in case you have time to add more tests.

The parts I didn't comment on seemed to have test coverage 🚀

PR #163, opened 2022-06-13
Fix an issue with a possible undefined contentWindow

alexandcote:


alxclark:
Seems like a reasonable fix, did you try creating a build in remote-ui and using it in a spin instance to test if it fixed the error ?

alxclark (approved):
This is still not clear to me if the iframe needs to release stuff before un-mounting and I'm not sure how to even test this, but testing the changes did fix the issue 👍🏼

lemonmade:
There is a `terminate` method you can call on an `endpoint`: https://github.com/Shopify/remote-ui/blob/main/packages/rpc/src/endpoint.ts#L139-L147. Right now, after you call that, we will still send `postMessage` (https://github.com/Shopify/remote-ui/blob/main/packages/rpc/src/endpoint.ts#L155). I feel like our current behavior is a bug, and we should stop sending messages after `terminate` is called. This would allow a consumer to prevent this bug from happening just by calling `terminate()` at the right time, and would allow all the adaptor libraries to continue to assume that they are always operating on an "alive" message channel.

packages/rpc/src/tests/endpoint.test.ts
```diff
@@ -116,5 +118,27 @@ describe('createEndpoint()', () => {
         message: expect.stringContaining('terminated'),
       });
     });
+
+    it('avoid sending post message on a terminated endpoint', () => {
```
lemonmade, line 122:
```suggestion
    it('avoids sending post message on a terminated endpoint', () => {
```

packages/rpc/src/tests/endpoint.test.ts
```diff
@@ -116,5 +118,27 @@ describe('createEndpoint()', () => {
...
+      const {port1} = new MessageChannel();
+      port1.start();
+
+      const messenger = fromMessagePort(port1);
+      const endpoint = createEndpoint(messenger, {createEncoder});
+
+      const spy = jest.spyOn(messenger, 'postMessage');
+
+      endpoint.terminate();
+      releaseRef!();
```
lemonmade, line 139: Hmm it's a bit confusing what this value actually refers to, I admit I am not even sure myself without digging into the `createEncoder` option a bit. It feels like it's testing an implementation detail, though. I think it would be nicer if we could have a "proper" example where a function is actually proxied across the bridge, and is then used to call the global `release()` function provided by the library

lemonmade (approved):

PR #164, opened 2022-06-14
Add failing test to demonstrate render order bug

ericfrank-stripe:
Yesterday, we upgraded to `@remote-ui/core` 2.1.11 and `@remote-ui/react` 4.5.3 (we were on the versions immediately before those, 2.1.10 and 4.5.2). Within a few hours, we had several developers report rendering bugs involving strange ordering and missing elements. This PR includes a test demonstrating one of the bugs introduced by the new version.

There's a detailed writeup with screenshots in an issue on our repo here: https://github.com/stripe/stripe-apps/issues/648

### DIFF
--- a/packages/react/src/tests/e2e.test.tsx
+++ b/packages/react/src/tests/e2e.test.tsx
@@ -555,4 +555,75 @@ describe('@remote-ui/react', () => {
     );
     expect(appElement.innerHTML).toBe('<img src="/image2.jpg">');
   });
+
+  it('renders the expected items after state updates', () => {
+    const RemoteBox = createRemoteReactComponent<
+      'Box',
+      {children?: any; id?: string; onClick?: () => void}
+    >('Box');
+
+    const HostBox = ({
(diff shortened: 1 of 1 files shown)


packages/react/src/tests/e2e.test.tsx
```diff
@@ -555,4 +555,75 @@ describe('@remote-ui/react', () => {
     );
     expect(appElement.innerHTML).toBe('<img src="/image2.jpg">');
   });
+
+  it.only('renders the expected items after state updates', () => {
```
cchalstrom-stripe, line 559: `only`❗

ericfrank-stripe: Fixed! Thanks

alexandcote:
Hi @ericfrank-stripe 

Thanks for reporting this issue. The fix should be in the latest version of `@remote-ui`. Keep me updated if you still experience the bug.

Thanks

PR #167, opened 2022-06-17
Add failing test showing nested fragment node bug

dannygarcia-stripe:
## Add failing tests

This adds a failing tests to showcase a possible bug. The test results in the following error after the simulated button click:

> Objects are not valid as a React child (found: object with keys {id, kind, children}). If you meant to render a collection of children, use an array instead.

This seems to happen when the react host is updating a nested fragment prop with a new component that hasn't yet been normalized. Therefore, it seems this new prop fails [this check](https://github.com/Shopify/remote-ui/blob/1eebeb255bad9391ae1bb028414a9580dde723dd/packages/react/src/host/RemoteComponent.tsx#L38) and passes through as a plain object (`{id, kind, children}` without `version`).

### DIFF
--- a/packages/react/src/tests/e2e.test.tsx
+++ b/packages/react/src/tests/e2e.test.tsx
@@ -741,4 +741,57 @@ describe('@remote-ui/react', () => {
     );
     expect(appElement.innerHTML).toBe('<img src="/image2.jpg">');
   });
+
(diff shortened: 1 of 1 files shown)


dannygarcia-stripe:
Closing in favor of #168

PR #168, opened 2022-06-27
Normalize new node props

dannygarcia-stripe:
Normalizes nodes with `addVersion` so newly rendered props with nested fragment children can render.

This will help us fix a user-reported issue: https://github.com/stripe/stripe-apps/issues/641


kumar303 (approved):
Thanks for the test 👍 Looks good from my limited understanding 😁

PR #170, opened 2022-07-12
Fix crash: reading children of undefined if in fragment

cchalstrom-stripe:
This fixes an error where we will try to read `[...currentInternals.children]`, but this is undefined in the case of a fragment.  This ends up resulting in a crash.

Added a test to cover this case.

cc @lemonmade


lemonmade (approved):
Another great find. Fix looks wonderful, and thank you as usual for the helpful test!

PR #172, opened 2022-08-11
Basic remote DOM

lemonmade:


marvinhagemeister:
FYI: @developit wrote a pretty complete DOM facade a few weeks back https://github.com/Shopify/remote-dom-experiment/blob/main/src/runtime/dom.ts

PR #173, opened 2022-08-12
rethrow errors to allow global error collection

robin-drexler:
Currently if an event handler throws an error, the error can't be listened to on the worker side because remote-ui swallows it and only later rejects the promise on the host side. (and vice versa)

This PR rethrows the original error which allows libraries like bugsnag and sentry to listen to these (in the form of `unhandledrejection`s in that case.


lemonmade (approved):
Wonderfully done 👍

PR #174, opened 2022-08-15
add changelog entry

robin-drexler:


mathieulag (approved):

lemonmade:
I would probably do a patch change for this, I don't think it was intentional to swallow the error.

robin-drexler:
@lemonmade I merged too soon. 😅 

Here's the PR that changes it to patch level https://github.com/Shopify/remote-ui/pull/175

PR #177, opened 2022-08-23
add allow-same-origin to demo iframe sandbox attribute

meirish:
Running the demo was erroring without this additional value in the `sandbox` attribute.


marvinhagemeister (approved):
Awesome, thanks for the PR 👍

PR #179, opened 2022-09-16
Fix react error-polyfill missing in dist files

marvinhagemeister:
@misha-reyzlin-shopify noticed that the changes from #178 weren't present in the packaged files. After a bit of debugging we noticed that `sideEffects: false` was set in `package.json` which drops any top-level code that is deemed to have side-effects. Simply removing that setting fixes it.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -29,7 +29,6 @@
       "require": "./host.js"
     }
   },
-  "sideEffects": false,
   "devDependencies": {
     "@types/react-dom": "^17.0.0",
     "react": "^17.0.0",


misha-reyzlin-shopify (approved):
👍

lemonmade:
Please change this to just list the files that actually have side effects, rather than disabling side effects altogether. This change means that the `@remote-ui/react/host` code can no longer be tree-shaken, even though it is entirely pure code.

marvinhagemeister:
@lemonmade sure can do that. Didn't know that it was possible to pass a list of files to `sideEffects`. Made a PR for that: https://github.com/Shopify/remote-ui/pull/180

PR #180, opened 2022-09-16
Add back sideEffect to only list effectful files

marvinhagemeister:
This fixes it in the same way as #179 .

See: https://github.com/Shopify/remote-ui/pull/179#issuecomment-1249545034

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -29,6 +29,9 @@
       "require": "./host.js"
     }
   },
+  "sideEffects": [
+    "./src/error-polyfill.ts"
+  ],
   "devDependencies": {
     "@types/react-dom": "^17.0.0",
     "react": "^17.0.0",


misha-reyzlin-shopify (approved):
🚀

lemonmade (commented):
I don't think this one will work when you ship it. I should have been more specific last time to say that we need to list all the *build* files that act as side-effects; the source files are not actually used by consumers. I think in this case it will be:

```json
{
  "sideEffects": [
    "./build/cjs/error-polyfill.cjs",
    "./build/esm/error-polyfill.mjs",
    "./build/esnext/error-polyfill.esnext",
  ]
}
```

marvinhagemeister:
@lemonmade listing the source file works. See: https://www.runpkg.com/?@remote-ui/react@4.5.10/build/cjs/render.js

PR #181, opened 2022-09-16
Add dist side-effectful files too

marvinhagemeister:
Note, the source file must still be present, otherwise the `build` command will strip it.


lemonmade (approved):
Thank you!

PR #186, opened 2022-10-21
Undo react patch that enabled error rethrowing in workers

misha-reyzlin-shopify:
This is done in favor of setting ErrorBoundary in the render function on the side of remote client which allows to listen to error on globalThis by reportError-ing in componentDidCatch

After this is merged and published we can merge and publish the respective PR (https://github.com/Shopify/checkout-web/pull/15073)  updating `packages/checkout-ui-extensions-react/src/render.tsx`


misha-reyzlin-shopify:
@lemonmade hey, good call and yes, we paired on the new solution as well!

PR #190, opened 2022-11-25
Allow custom React reconciler in remote-ui/react

alxclark:
## What is this solving ?

In some cases, it's possible to have a remote-ui setup where there is two primary renderers on the remote, one for the main content on the remote and one for the components being rendered and sent to the host.

With this kind of setup, React will throw an error when the same context provider is used with both renderers. This is due to the fact that React only supports having one primary renderer and one secondary renderer concurrently. Some unexpected behaviour can occur from this and React will also log warnings to the console.

Here's a reproduction of this exact issue using an iframe which renders content and also attempts to render components on the host (the main webpage here). The shared context provider here is a `RecoilRoot` from [recoil](https://github.com/facebookexperimental/Recoil) and is used only as an example of a library that could be used in more than one environment.

https://stackblitz.com/edit/vitejs-vite-qmh2qb?file=src%2Fremote.tsx

image

## How is it solving the issue ?

In cases like this, we need to tell remote-ui's reconciler that it is being used as a secondary renderer. React's reconciler package accepts an option called [isPrimaryRenderer](https://github.com/facebook/react/tree/main/packages/react-reconciler#isprimaryrenderer) that allows us to do just that.

In order to let consumers customize their React reconciler as they see fit, this PR does the following:
* Add a `reconciler` parameter to `render` (remote-ui/react)
* Expose a `createReconciler` function to create a reconciler which defaults to our existing configuration.

Ultimately, this allows consumers to create a reconciler like this:

```tsx
const reconciler = createReconciler({isPrimaryRenderer: false});

// and eventually use the reconciler
render(<>{children}</>, root, root.mount, reconciler);
```

## Notes
* There is no test coverage around the React reconciler, so I haven't added extra test for this change which is pretty straightforward. Happy to look into it if we are worried about the change.
* I'm not sure how much flexibility we want to expose to consumers. I've exposed the full host config for developers to customize, but we might want to restrict it. I'm all ears for feedback and suggestions on this.

### DIFF
--- a/packages/react/src/reconciler.ts
+++ b/packages/react/src/reconciler.ts
@@ -1,4 +1,7 @@
-import reactReconciler, {Reconciler as ReactReconciler} from 'react-reconciler';
+import reactReconciler, {
+  Reconciler as ReactReconciler,
+  HostConfig,
+} from 'react-reconciler';
 import type {
   RemoteRoot,
   RemoteText,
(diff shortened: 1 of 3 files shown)


packages/react/src/reconciler.ts
```diff
@@ -27,148 +30,166 @@ export type Reconciler = ReactReconciler<
...
-  },
-  prepareUpdate(_instance, _type, oldProps, newProps) {
-    const updateProps: Record<string, unknown> = {};
-    let needsUpdate = false;
-
-    for (const key in oldProps) {
-      if (!has(oldProps, key) || key === 'children') {
-        continue;
+export const createReconciler = (
+  hostConfig?: HostConfig<
```
lemonmade, line 34: Can we make this a more specific options object, that is just `options?: {primary?: boolean}`? I don't think we need to leave space for overriding the rest of the options here.

lemonmade (approved):
Thanks for the addition! I do find this particular way of using remote-ui to be a little brain-melting but I don't see any harm in making it possible!

Issue #184, opened 2022-10-04
Error in iframe-parent: null is not an object (evaluating 'e.contentWindow.postMessage')


kumar303:
In Shopify's production bug tracker, we're seeing a high frequency of an error at [this line](https://github.com/Shopify/remote-ui/blob/66b10f13a04077681888789f6215ae75350c3a6e/packages/rpc/src/adaptors/iframe-parent.ts#L34):

```
TypeError null is not an object (evaluating 'e.contentWindow.postMessage') 
    node_modules/@remote-ui/rpc/build/esnext/adaptors/iframe-parent.esnext:28:27 
    [native code] asyncFunctionResume
    [native code] 
    [native code] promiseReactionJobWithoutPromise
```

It happens in these browsers: Mobile Safari, Mobile Safari UI/WKWebView, Chrome Mobile iOS, Safari.

I'm not sure if it's a compat issue or a cross domain issue.

kumar303:
I don't think we updated `remote-ui` recently. @jamesvidler do you know?

marvinhagemeister:
Just checked: The last time we updated `remote-ui` was about 3 weeks ago, way before #183 .

kumar303:
Thanks for these detailed notes 👍 

> Semi-related to this, the `iframeReadyPromise` has a bug where it will continuously listen to `message` events even after the promise is resolved/the iframe is removed. This causes a memory leak.

I see it [here](https://github.com/Shopify/remote-ui/blob/89f838840f7bd54f2e09cbe31276e84d634db1f0/packages/rpc/src/adaptors/iframe-parent.ts#L21), too. Good catch. Can you file a separate issue for that?

TheCloudlessSky:
@kumar303 done in #192!

PR #192, opened 2022-11-29
Fix memory leak when listening to 'message' events from child iframes

TheCloudlessSky:
Related to #184. `fromIframe` message endpoint has two bugs:

1. It will listen to `message` events even after the promise is resolved or the iframe is removed. This will continuously compound as iframes are added/removed from the DOM.
2. It's also possible that the `remote-ui::ready` message is never received (e.g. the iframe fails to lead), so the event handler is never cleaned up.

This PR addresses both issues to avoid memory leaks.

I didn't address testing in this PR, but happy to add as directed 😄


TheCloudlessSky:
@lemonmade done!

lemonmade:
Congratulations, you are the proud new owner of the very first Changesets-based release of remote-ui, with [`@remote-ui/rpc@1.3.5`](https://github.com/Shopify/remote-ui/releases/tag/%40remote-ui%2Frpc%401.3.5) 😛 Thank you again for the great fix!

TheCloudlessSky:
Awesome - glad to help! 💅

PR #205, opened 2023-02-09
docs: Clarify react and react-reconciler dep requirements by version

banderson:
### Context:

I noticed what seemed to be a typo about the React peer dependency version requirements in the README, referencing 17.x where it should be 18.x. While updating that, I also noticed the install commands for the latest version `@5.0.0` weren't specifying a specific dep versions, so I added those too. 

I'm not sure if this is helpful or not, so disregard if not :)

### Changes included:

 - Update documented dependency version requirements for `react` and `react-reconciler` for latest v5 release


lemonmade (approved):
Thanks so much @banderson, these corrections are definitely needed 👍

PR #213, opened 2023-02-21
add vanilla ts vite example

robin-drexler:
Adds a vitejs powered version of the vanillajs example.
It's a bit of a prerequisite to work on https://github.com/Shopify/remote-ui/pull/207#pullrequestreview-1299795163

It'll be much easier to work on stuff if it can be checked in a browser without much setup.

Changes to remote-ui packages are picked up right away. 

🎩 
- checkout this branch
- run `yarn`
- `cd examples/vanilla-dom-vite`
- `yarn dev --open`
- type into the textfield, click the button and see your message being logged

@lemonmade Do you think this could be useful? 
I'd like to add a react version and one that uses a worker to communicate instead of an iframe. I do wonder tho if we should try to put it all in one vite app or if it'd be better to create one for each use-case. 🤔 

Wdyt?


examples/vanilla-dom-vite/app/types.ts
```diff
@@ -0,0 +1,8 @@
+import { type RemoteChannel } from "@remote-ui/core";
+
+export type Api = {
+  getMessage(): Promise<string>;
+}
+export type EndpointApi = {
+  render(receiver: RemoteChannel, api: Api): Promise<unknown>
+}
```
lemonmade, line 8: Nit: missing newline

examples/vanilla-dom-vite/vite.config.js
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export default defineConfig({
+  root: 'app',
+  resolve: {
+    alias: {
+      '@remote-ui/core': getPathForPackage('core'),
+      '@remote-ui/rpc': getPathForPackage('rpc'),
+      '@remote-ui/dom': getPathForPackage('dom'),
+    },
```
lemonmade, line 17: Vite doesn't do this automatically, based on the contents of the `tsconfig.json`?

lemonmade (approved):
Looks great 🚀

examples/vanilla-dom-vite/vite.config.js
robin-drexler: I haven't set up path aliases in the tsconfig. I could try and see if it fixes it.

robin-drexler: looks like folks usually do aliases in both. 
https://www.npmjs.com/package/vite-tsconfig-paths seems to address this, but I think it's fine for these few packages?

PR #215, opened 2023-02-23
ensure child iframe message is not lost

robin-drexler:
Fixes https://github.com/Shopify/remote-ui/pull/207, but (hopefully) properly.

parent iframe now also sends a message once it's ready to receive messages which will trigger the child to resend its ready message, which will in turn resolve the parent's promise. If the child's document isn't ready when the parent sends its message, it'll be discarded, but it'll be sent once the child's document is ready. At this point the parent is already ready to receive messages as it has already sent a message on its own. 😅 

First commit has the actual changes. Second commit adjusts the vite example app to use the html iframe again + formats the ts file. 😬 

🎩 

- Run the vite app and ensure the button shows up. `yarn run example:vanilla-dom-vite`
- Once the package is released, the non-vite vanilla app should also work reliably again. 

I've also created a [`iframe-fix-tophat` branch](https://github.com/Shopify/remote-ui/compare/main...robin-drexler:remote-ui:iframe-fix-tophat?expand=1) where the endpoint setup in the frame is delayed by 2 seconds. After those two seconds the button shows up, which isn't the case with the state in main.

### DIFF
--- a/.changeset/perfect-students-pull.md
+++ b/.changeset/perfect-students-pull.md
@@ -1,0 +1,5 @@
+---
+'@remote-ui/rpc': patch
+---
+
+Ensures iframe connection can always be established


mawburn (commented):
@lemonmade pointed me here because I had a similar idea based on some recent issues we ran in to. 

What would be very nice to have in addition to this, is an optional timeout for the `ready` event that will throw an error if the iframe isn't ready within a certain amount of time. 

While this gives us an event when the iframe becomes ready, we are still in the same problem that we are facing where it will wait forever if it never becomes ready.

robin-drexler:
Interesting. I assume it'd be sufficient to only add this to the `fromIframe` function? Not sure how useful it'd be inside the iframe?

mawburn:
Ok, apologies. After digging deeper into this, this is solving a different problem than what I originally thought. The problem I was referring to is going the other direction where the parent is constantly waiting on the child because the child never loads. I will see about putting up a PR for the changes I talked about in my other comment.

packages/rpc/src/adaptors/iframe-child.ts
```diff
@@ -7,12 +7,31 @@ export function fromInsideIframe({targetOrigin = '*'} = {}): MessageEndpoint {
     );
   }
 
+  /**
```
lemonmade, line 10: Nit: these aren't documentation, so we generally use single-line comments for them instead.

packages/rpc/src/adaptors/iframe-child.ts
```diff
@@ -7,12 +7,31 @@ export function fromInsideIframe({targetOrigin = '*'} = {}): MessageEndpoint {
...
+   */
+
   const {parent} = self;
 
   const ready = () => parent.postMessage('remote-ui::ready', targetOrigin);
 
-  // Listening to `readyState` in iframe, though the child iframe could probably
-  // send a `postMessage` that it is ready to receive messages sooner than that.
+  window.addEventListener('message', (event) => {
+    if (event.source !== parent || document.readyState !== 'complete') {
```
lemonmade, line 26: Do you know if it's possible for this to happen for the document is ready?

packages/rpc/src/adaptors/iframe-child.ts
```diff
@@ -7,12 +7,31 @@ export function fromInsideIframe({targetOrigin = '*'} = {}): MessageEndpoint {
...
+      return;
+    }
+
+    if (event.data === 'remote-ui::ready') {
```
lemonmade, line 30: Can we pull this out into a constant?

lemonmade (changes requested):

packages/rpc/src/adaptors/iframe-child.ts
robin-drexler: Do you mean _before_ the document is ready?

robin-drexler: I actually had the same question, but assumed that the previous authors have checked this. 

But I'd try to check

robin-drexler: @lemonmade it seems that this can actually happen. It was hard to reproduce since all our test apps use `type=module` which defers scripts and hence only executes them after the document is loaded.

However, if that's not the case, `document.readyState` can be `loading` under some circumstances when the iframe loads a blocking JS after the file that contains the rpc code.

image

lemonmade (approved):
Looks great, thanks for working on this! [Add a changeset](https://github.com/Shopify/remote-ui/blob/main/CONTRIBUTING.md#releasing-changes) and you are good to go 👍

PR #217, opened 2023-03-03
introduce react worker example

robin-drexler:
Adds react worker example to the vite app and also renames it to be a kitchen sink rather than just the vanilla example.

I think it makes sense to just put all those local apps into a single vite-based project.

The index page links to both currently existing example apps.

image
image

🎩 
- checkout this branch
- `yarn`
- `yarn example:kitchen-sink-vite`
- open the website as mentioned in the terminal output


examples/kitchen-sink-vite/app/react-worker/host/remote-renderer.tsx
```diff
@@ -0,0 +1,35 @@
...
+  createRemoteReceiver,
+  RemoteRenderer as OriginalRemoteRenderer,
+} from '@remote-ui/react/host';
+
+import {createEndpoint, fromWebWorker} from '@remote-ui/rpc';
+
+import {Button} from './components';
+import {Endpoint} from '../types';
+
+const remoteEndpoint = createEndpoint<Endpoint>(fromWebWorker(new Worker()));
```
lemonmade, line 15: Should we do this as part of the effect instead? Most React apps should not create a worker in this way, since it prevents SSR.

examples/kitchen-sink-vite/app/react-worker/host/remote-renderer.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export function RemoteRenderer({
```
lemonmade, line 17: I think we should find a different name for this to avoid conflicting with the library's component name.

examples/kitchen-sink-vite/app/index.html
```diff
@@ -0,0 +1,24 @@
...
+    <title>remote-ui kitchen sync</title>
+    <link rel="icon" href="https://fav.farm/🛰️" />
+    <link rel="stylesheet" href="./style.css" />
+  </head>
+  <body>
+    <main>
+      <h1>remote-ui kitchen sink</h1>
+      <ul>
+        <li>
+          <a href="/vanilla-iframe/">Vanilla iframe</a>
```
lemonmade, line 16: Can we add a little extra styling to make these list items nicer? Removing the default list styles + larger tap targets would be great.

PR #223, opened 2023-04-03
Fix TypeScript types for exports consumers

lemonmade:
TypeScript consumers using remote-ui packages and the new [`'nodenext'` module resolution strategy](https://www.typescriptlang.org/tsconfig/#moduleResolution) get the following error:

Screenshot 2023-04-03 at 9 15 14 AM

This is because we have `exports`, but do not define `types` conditions, and TypeScript refuses to load our "fallback" types instead.

To fix this, I define the `types` condition for `exports`. Older TypeScript/ non-`nodenext` consumers should not be affected by this option, but newer TypeScript using `nodenext` get types again.


lemonmade:
I ended up bumping the project to use TypeScript 5.0 in order to solve some problems with type checking.

PR #226, opened 2023-04-06
Exposing types definitions in core package

alexandcote:
Introduced in: https://github.com/Shopify/remote-ui/pull/226

Exposing `types` definition in `@remote-ui/core`.

Fixing an issue: 
```
Could not find a declaration file for module '@remote-ui/core'. '/app/node_modules/@remote-ui/core/index.js' implicitly has an 'any' type.
--
  | Try `npm i --save-dev @types/remote-ui__core` if it exists or add a new declaration (.d.ts) file containing `declare module '@remote-ui/core';`


```

### DIFF
--- a/.changeset/healthy-crews-build.md
+++ b/.changeset/healthy-crews-build.md
@@ -1,0 +1,5 @@
+---
+'@remote-ui/core': patch
+---
+
+Re-exposing types definitions in @remote-ui/core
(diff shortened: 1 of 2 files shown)


lemonmade (approved):
Whoops, thanks for catching that and putting up a fix!

PR #242, opened 2023-09-21
Check for target contentWindow before posting a message

### LINKED ISSUE #184: Error in iframe-parent: null is not an object (evaluating 'e.contentWindow.postMessage')

mikewilson01:
This attempts to fix #184 by checking to see if the `target.contentWindow` is still defined before attempting to post a message.

We're seeing a lot of this error in checkout, and it seems to be related to the installments iframe. What is happening is that we're closing the iframe modal, and in a `useEffect` return, we're calling `terminate()` on the endpoint. However, at this point, React has already removed the iframe from the DOM, so `target.contentWindow` is undefined.

If this feels like too broad of a fix, another option is to try and narrow this to just `TERMINATE` message?

i.e.

```ts
// ...
async postMessage(message, transfer) {
  await iframeReadyPromise;

  if (isTerminateMessage(message) && !target.contentWindow) {
    return;
  }

  target.contentWindow.postMessage(message, targetOrigin, transfer);
}

//...


function isTerminateMessage(message) {
  if (Array.isArray(message) && message.length === 1) {
    if (message[0] === TERMINATE) {
      return true;
    }
  }

  return false;
}

### CODE REVIEW

File: packages/rpc/src/adaptors/iframe-parent.ts [hunk 1]
```diff
@@ -41,6 +41,11 @@ export function fromIframe(
   return {
     async postMessage(message, transfer) {
       await iframeReadyPromise;
+
+      if (!target.contentWindow) {
```
henrytao-me, line 45: We could do `target.contentWindow?.postMessage` like we do in line 34.

lemonmade (approved):
Makes sense to me, good find!

PR #246, opened 2023-10-23
Handle callId being potentially undefined

BPScott:
Hopeful fix for Shopify/checkout-web#26840.

I've not proven this fixes the issue, but it seems likely.

We were using a non-null assertion to claim that the `callId` could never be undefined, however in the real world we've been seeing lots of cases where it is undefined. So handle that at runtime using optional chaining.

### DIFF
--- a/.changeset/giant-points-double.md
+++ b/.changeset/giant-points-double.md
@@ -1,0 +1,5 @@
+---
+'@remote-ui/rpc': patch
+---
+
+Handle cases where callId in results may be undefined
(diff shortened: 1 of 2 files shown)


frehner:
> It would be nicer to figure out why `callIdsToResolver.get(callId)` is undefined / empty 🤔

Partial explanation here, though not really the root cause. 

https://github.com/Shopify/checkout-web/issues/26840#issuecomment-1760109442

BPScott:
Closing this. Original error was due to somebody calling `createEndpointFromIFrame` multiple times on a single iframe (https://github.com/Shopify/checkout-web/pull/27800)

PR #249, opened 2023-10-30
Update documentation links

antoinegrant:
Fixed the documentation links to point to the root instead of relative to the package.


lemonmade (approved):
Thanks Antoine!

PR #253, opened 2024-01-11
Fix parameter type for Node.trigger

danpaz:
While writing tests for a Checkout UI extension, I noticed the type signature for `trigger` seemed to be incorrect as it was inferring the function's return type as the function's parameters type. This may have been a copy-paste error from https://github.com/Shopify/remote-ui/pull/223/files#diff-0eca05e10e9cb7143453a3be68ba1617a04bb006cf58988be21b759143d4c339R73-R76.

I tested this locally with `yarn link` and verified the type inference is now working correctly.

### DIFF
--- a/.changeset/shaggy-ladybugs-hang.md
+++ b/.changeset/shaggy-ladybugs-hang.md
@@ -1,0 +1,5 @@
+---
+'@remote-ui/testing': patch
+---
+
+Fixes parameter type for Node.trigger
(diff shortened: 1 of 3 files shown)


danpaz:
I have signed the CLA!

lemonmade (approved):
Sorry for the long delay, and thank you for the fix! The CHANGELOG entries are created automatically, so I will delete the one you've added and get the new version published.

PR #255, opened 2024-02-09
Allow `File` (and other object types) to be send over RPC

robin-drexler:
This is a request for comment PR, @lemonmade 

The main intention was to allow `File` to be send over the RPC layer. 

Before we'd try to decode individual properties of the object, essentially destroying it.

Now we're only doing this for basic objects, similar to how it's done in `encode`.

This basically allows any [supported value](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) to be send as is now.

I'm not sure if that introduces unintended side-effects tho, hence the request for comments because I know we had issues with infinite loops prior.

Alternatively, and if we want to play it safe, we'd also just add another exception for `File`, like we already had one for `ArrayBuffer`.

## 🎩 

I tested this down to Safari 13. In earlier versions, even the vite prod build would produce JS that safari doesn't understand and I didn't wanna set up the legacy build. 

- check out `rd/file-support-tophat` (which is another branch that has some more changes to tophat this
- run `yarn example:kitchen-sink-vite`
- Go to http://localhost:5173/
- Click on React worker
- Open console
- Select a file. Preferably an image or something a browser can display
- A blob url should be logged in the console. That is being logged from the worker
- Copy and paste it into your address bar
- The file should render

image

image


packages/rpc/src/encoding/basic.ts
```diff
@@ -209,13 +205,15 @@ export function createBasicEncoder(api: EncodingStrategyApi): EncodingStrategy {
...
       }
 
-      return Object.keys(value).reduce(
-        (object, key) => ({
-          ...object,
-          [key]: decode((value as any)[key], retainedBy),
-        }),
-        {},
-      ) as any;
+      if (isBasicObject(value)) {
```
robin-drexler, line 208: If it's not basic, it'll be returned as is further down

robin-drexler: Not sure if there's "non basic" cases, where we'd need to iterate over properties.

lemonmade (approved):
Thanks Robin, LGTM! This is the intended behavior, so I think the current behavior is a bug. Can you add a changeset for a patch change?

PR #400, opened 2024-07-26
Add breadcrumbs to remote-ui internals, and accept a logging callback

DanielParpal:
**Context**

It can be difficult to troubleshoot prod issues related to the inner workings of remote-ui/dom, especially when such issues are difficult to reproduce locally (eg. webkit-specific or timing issues) and originate from the sandbox.

**Suggestion**

This PR suggests to add breadcrumbs throughout the `createEndpoint` lifecycle, and allow clients of this library to retrieve such breadcrumbs through a `logCallback` optional callback. It takes advantage of the existing error messages passing from one endpoint to another and adds an optional `breadcrumbs` property to the error-like object. 

Thus, whenever `waitForResult` notices an error, it calls `logCallback` (if defined) before rejecting the promise.

Example of a very basic `remote-ui `client setup:
```tsx
    endpointRef.current = createEndpoint(
      fromIframe(iframeRef.current, { targetOrigin: '*' }),
      {
        logCallback(breadcrumbs) {
          if (breadcrumbs) {
            setSandboxBreadcrumbs(breadcrumbs);
          }
        },
      },
    );
```

And a very simple example of an app using this feature to append the breadcrumbs to the DOM when an error is thrown in the sandbox:

https://github.com/user-attachments/assets/48ff2009-1bb5-4b28-8cb5-b4d64b8c3228

### DIFF
--- a/packages/rpc/src/endpoint.ts
+++ b/packages/rpc/src/endpoint.ts
@@ -15,21 +15,28 @@ export const RELEASE = 3;
 export const FUNCTION_APPLY = 5;
 export const FUNCTION_RESULT = 6;
 
+const BREADCRUMB_PREFIX = 'remote-ui';
+
 type AnyFunction = (...args: any[]) => any;
+
+interface ErrorWithBreadcrumbs extends Error {
+  breadcrumbs?: string;
+}
 
 interface MessageMap {
   [CALL]: [string, string | number, any];
-  [RESULT]: [string, Error?, any?];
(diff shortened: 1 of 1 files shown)


lemonmade (changes requested):
@DanielParpal sorry, but I think these additions are too specific and narrowly useful to add to the RPC library. I can see how they would be useful for debugging a particular issue (e.g., building the library with this debugging and using it in a local project definitely makes sense), but you're instrumenting a part of the library that is intended to be bug-free, so I don't think it will be useful more generally.

PR #401, opened 2024-08-01
Add missing document elements

olavoasantos:
## Summary

The [HTML `Document` interface](https://developer.mozilla.org/en-US/docs/Web/API/Document#instance_properties) expects the instance to have a [body](https://developer.mozilla.org/en-US/docs/Web/API/Document/body), [head](https://developer.mozilla.org/en-US/docs/Web/API/Document/head) and [documentElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement) element attached to it which were missing from the polyfill. This PR adds those properties to the class.

### DIFF
--- a/packages/polyfill/source/Document.ts
+++ b/packages/polyfill/source/Document.ts
@@ -12,6 +12,9 @@ import {Comment} from './Comment.ts';
 import {DocumentFragment} from './DocumentFragment.ts';
 import {HTMLTemplateElement} from './HTMLTemplateElement.ts';
 import {isParentNode, cloneNode} from './shared.ts';
+import {HTMLBodyElement} from './HTMLBodyElement.ts';
+import {HTMLHeadElement} from './HTMLHeadElement.ts';
+import {HTMLHtmlElement} from './HTMLHtmlElement.ts';
 
 export class Document extends ParentNode {
   nodeType = NodeType.DOCUMENT_NODE;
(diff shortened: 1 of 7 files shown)


packages/polyfill/source/Element.ts
```diff
@@ -64,6 +64,10 @@ export class Element extends ParentNode {
     return sib;
   }
 
+  constructor(name: string = '') {
```
vividviolet, line 67: Why do we need this fallback?

packages/polyfill/source/Document.ts
```diff
@@ -82,7 +91,7 @@ export function createElement<T extends Element>(
   } else if (lowerName === 'template') {
     element = new HTMLTemplateElement() as any;
   } else {
-    const CustomElement = document.defaultView!.customElements.get(name);
+    const CustomElement = ownerDocument.defaultView.customElements.get(name);
```
vividviolet, line 94: Why don't we pass `name` to the constructor?

vividviolet (commented):
Looks good to me. I just have a few questions

packages/polyfill/source/Element.ts
olavoasantos: I kept the behaviour that existed before. I can remove it and enforce the element name as a requirement.

packages/polyfill/source/Document.ts
olavoasantos: I'm not sure. It was the behaviour that is implemented.

packages/polyfill/source/Document.ts
```diff
@@ -11,14 +11,28 @@ import {Comment} from './Comment.ts';
...
+import {HTMLBodyElement} from './HTMLBodyElement.ts';
+import {HTMLHeadElement} from './HTMLHeadElement.ts';
+import {HTMLHtmlElement} from './HTMLHtmlElement.ts';
 
 export class Document extends ParentNode {
   nodeType = NodeType.DOCUMENT_NODE;
   [NAME] = '#document';
+  body: Element;
+  head: Element;
+  documentElement: Element;
```
lemonmade, line 23: Should we type these as their more specific subclasses?

lemonmade (approved):
Looking good, thanks Olavo! Just one question on the typing of the new fields.

PR #402, opened 2024-08-02
Move hooks into Window instance

olavoasantos:
## Summary

The current implementation of the DOM polyfill uses an in memory singleton to manage internal hooks. This results in coupling of the polyfill with the global scope, preventing multiple instances of Window from being created without interfering with each other.

This PR refactors the management of these hooks by scoping them in the Window instance, decoupling the implementation from the global scope.

### DIFF
--- a/packages/polyfill/source/CharacterData.ts
+++ b/packages/polyfill/source/CharacterData.ts
@@ -6,7 +6,7 @@ export class CharacterData extends ChildNode {
 
   constructor(data: any) {
     super();
-    this[DATA] = data != null ? String(data) : '';
+    this[DATA] = data == null ? '' : String(data);
   }
 
   protected setData(data: any) {


packages/polyfill/source/Attr.ts
```diff
@@ -45,7 +45,12 @@ export class Attr extends Node {
     this[VALUE] = str;
     const ownerElement = this[OWNER_ELEMENT];
     if (!ownerElement) return;
-    hooks.setAttribute?.(ownerElement as any, this[NAME], str, this[NS]);
+    this.ownerElement?.ownerDocument.defaultView[HOOKS].setAttribute?.(
```
vividviolet, line 48: Can we get away with doing `this[OWNER_DOCUMENT]` or does that not work?

.changeset/chilled-phones-explain.md
```diff
@@ -0,0 +1,6 @@
+---
+'@remote-dom/polyfill': patch
+'@remote-dom/core': patch
```
lemonmade, line 3: I think I'd consider this a minor change for both. It does give you a capability you didn't have before, even if the default experience is implemented identically before and after the change.

packages/polyfill/source/CharacterData.ts
```diff
@@ -16,7 +15,13 @@ export class CharacterData extends ChildNode {
       str = typeof data === 'string' ? data : String(data);
     }
     this[DATA] = str;
-    hooks.setText?.(this as any, str);
+    if (this[OWNER_DOCUMENT]) {
+      this[OWNER_DOCUMENT].defaultView[HOOKS].setText?.(this as any, str);
+    } else {
+      setTimeout(() =>
```
lemonmade, line 21: This is an issue because of us calling `setData` in the constructor, right? Instead of doing this, could we just assign to `this[DATA]` directly in the constructor (applying the same stringifying logic implemented above), which avoids this triggering a hook call?

packages/polyfill/source/NamedNodeMap.ts
```diff
@@ -60,7 +67,11 @@ export class NamedNodeMap {
         if (prev) prev[NEXT] = attr[NEXT];
         if (this[CHILD] === attr) this[CHILD] = attr[NEXT];
         updateElementAttribute(ownerElement, attr.name, attr.value, null);
-        hooks.removeAttribute?.(ownerElement as any, name, namespaceURI);
+        ownerElement[OWNER_DOCUMENT].defaultView[HOOKS].removeAttribute?.(
```
lemonmade, line 70: What do you think about adding a `HOOKS` getter to `Node` as well, that internally does the `[OWNER_DOCUMENT].defaultView[HOOKS]` accessing?

packages/polyfill/source/EventTarget.ts
```diff
@@ -1,14 +1,15 @@
...
 import {CAPTURE_MARKER, type Event} from './Event.ts';
 import type {ChildNode} from './ChildNode.ts';
+import type {Document} from './Document.ts';
 
 const ONCE_LISTENERS = Symbol('onceListeners');
 
 export class EventTarget {
   [LISTENERS]?: Map<string, Set<EventListenerOrEventListenerObject>>;
   [ONCE_LISTENERS]?: WeakMap<EventListenerOrEventListenerObject, EventListener>;
+  [OWNER_DOCUMENT]?: Document;
```
lemonmade, line 12: Can we make this `private`? It might also be worth mentioning in a comment how this value is set (IIUC, it is only set on some subclasses, and we are relying on that behavior here).

lemonmade (changes requested):
I like this change a lot, thanks for putting it together. I left a few comments but I think this is quite close 👍

packages/polyfill/source/CharacterData.ts
olavoasantos: I considered that but that would skip the initial hook call and change the existing behaviour. I'm ok with that but it's a change in the existing behaviour. Alternatively, we can manually set the value via `this[DATA]` and schedule the hook call within the constructor.

packages/polyfill/source/NamedNodeMap.ts
olavoasantos: That's a good call. Should clean things up.

packages/polyfill/source/EventTarget.ts
olavoasantos: That's fair. IIRC these keywords are only for TypeScript and shouldn't have any effect at runtime, but I'm ok with adding it. In this case, it might be `protected` instead of `private` as these are inherited.

.changeset/chilled-phones-explain.md
olavoasantos: That makes sense. I'll make them a minor

packages/polyfill/source/CharacterData.ts
lemonmade: I am OK to break this behavior. I think the library is actually broken in a different way here — we expose a `createText` hook https://github.com/Shopify/remote-dom/blob/ce50f4bf86f59cbf0a33cfa33144d47125131a3e/packages/polyfill/source/hooks.ts#L10, which should be used to cover the "initial value of a text node" case, but we don't seem to call that hook anywhere 😬

olavoasantos: Whoops haha Should I add `setTimeout(() => this[HOOKS]?.createText ...)` here then?

PR #406, opened 2024-08-08
fix(polyfill): dedupe hooks.addEventListener calls

developit:
This fixes `hooks.addEventListener()` being called even when EventTarget.addEventListener rejects a duplicate listener registration, allowing handler counting to be implemented without maintaining a separate Set of listeners.

The corresponding behavior was already implemented in `removeEventListener`, so I think this can just be considered a bugfix and landed without the need for any new hook.


lemonmade:
Thanks Jason!

PR #465, opened 2024-10-10
Fix `slot` to be transmitted as an attribute, not a property

lemonmade:
While debugging Customer Accounts with @robin-drexler, we noticed that the React host utilities were expecting the `slot` of an element to be available in its `attributes` https://github.com/Shopify/remote-dom/blob/main/packages/react/source/host/hooks/props-for-element.tsx#L85, but the core library was actually sending it as a property instead. The Preact version was looking at the property as well, so this issue was hidden when playing with the comprehensive example, which uses Preact on the host (even when using React for the example).

To fix this, I am aligning everything to send/ read `slot` as an attribute, not a property.

### DIFF
--- a/packages/core/source/elements/RemoteElement.ts
+++ b/packages/core/source/elements/RemoteElement.ts
@@ -643,7 +643,7 @@ export abstract class RemoteElement<
       attribute === 'slot' &&
       (this.constructor as typeof RemoteElement).slottable
     ) {
-      updateRemoteElementProperty(
+      updateRemoteElementAttribute(
         this,
         attribute,
         newValue ? String(newValue) : undefined,
--- a/.changeset/pink-dryers-fix.md
+++ b/.changeset/pink-dryers-fix.md
@@ -1,0 +1,6 @@
+---
+'@remote-dom/preact': patch
+'@remote-dom/core': patch
+---
+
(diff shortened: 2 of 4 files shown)


packages/core/source/elements/RemoteElement.ts
```diff
@@ -643,7 +643,7 @@ export abstract class RemoteElement<
       attribute === 'slot' &&
       (this.constructor as typeof RemoteElement).slottable
```
vividviolet, line 644: Do we need this guard in the property setter? If not, would we get into a situation where the component implementation reflects the attribute even though it's not supported?

vividviolet (commented):
The change makes sense to me. Just have a question

packages/core/source/elements/RemoteElement.ts
lemonmade: I don't think we want to prevent a `slot` attribute/ property from getting written in the same way they would be for the DOM. This additional, `RemoteElement`-specific `slottable` property is just for deciding whether that attribute should be synchronized with the host. I kind of wish I had not made this optional at all, since I can't imagine a reason why you would not want to send `slot` to the host, especially since the host needs to guard against unexpected attributes anyways.

PR #472, opened 2024-10-22
Fix references after node removal

jakearchibald:
Prior to this:

- `parentNode`
- `nextSibling`
- `previousSibling`
- `nextElementSibling`
- `previousElementSibling`

…remained set when removing an element from its parent.

I also added `node.parentElement`.


vividviolet (approved):

henrytao-me (approved):
🎩-ed. LGTM

Oleksii20in (commented):
😁

PR #485, opened 2024-11-05
Make react example use react

igor10k:
`@preact/preset-vite` by default aliases `react` to `preact` so react example in kitchen sink was actually using preact under the hood.


simontaisne (approved):
LGTM! Makes it closer to a real-life example 👍

PR #492, opened 2024-11-14
Apply React polyfill directly to `globalThis`

lemonmade:
This PR updates the React polyfill to apply directly to `globalThis`. This fixes a bug that resulted from shipping https://github.com/Shopify/remote-dom/pull/470, since the additional polyfills added by `@remote-dom/react/polyfill` (in particular, the one for `HTMLIframeElement`) were only being applied to the polyfilled `Window` class, not to `globalThis`.

### DIFF
--- a/.changeset/sharp-frogs-bathe.md
+++ b/.changeset/sharp-frogs-bathe.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/react': patch
+---
+
+Apply React polyfill directly to `globalThis`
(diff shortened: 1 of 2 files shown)


robin-drexler (approved):

lemonmade:
I've updated this PR to depend on the changes in https://github.com/Shopify/remote-dom/pull/499, where I rolled back the `globalThis === globalThis.window === globalThis.self` invariant.

PR #498, opened 2024-11-19
test snapit

robin-drexler:


simontaisne:
@robin-drexler snapit is running into `E404 Not Found` errors when trying to publish the `remote-dom/core`, `@remote-dom/preact`, and `@remote-dom/signals` packages to NPM. Do you think it might be because we're missing the `NPM_TOKEN` environment variable?

https://github.com/Shopify/remote-dom/blob/d22c88e07bcacefaa9c93d54141aa4324be4401d/.github/workflows/snapit.yml#L22-L23

robin-drexler:
> @robin-drexler snapit is running into `E404 Not Found` errors when trying to publish the `remote-dom/core`, `@remote-dom/preact`, and `@remote-dom/signals` packages to NPM. Do you think it might be because we're missing the `NPM_TOKEN` environment variable?
> 
> https://github.com/Shopify/remote-dom/blob/d22c88e07bcacefaa9c93d54141aa4324be4401d/.github/workflows/snapit.yml#L22-L23

yeah it doesn't work (yet)

We had to merge the action to actually try it. 

I'm not sure yet why it's not working, but noticed the 404.

We do feed it a token: https://github.com/Shopify/remote-dom/blob/main/.github/workflows/snapit.yml#L23, but maybe it's not the correct one. 
It's the same one we use in [other places](https://github.com/Shopify/remote-dom/blob/d22c88e07bcacefaa9c93d54141aa4324be4401d/.github/workflows/deploy.yml#L28), but as `NODE_AUTH_TOKEN`.

github actions are unfortunately quite difficult to test.

Issue #488, opened 2024-11-13
Polyfill conflicts with ThreadWebWorker


j4m3sb0mb:
It seems that `@remote-dom/core/polyfill` is conflicting with `ThreadWebWorker` from `@quilted/threads` in a WebWorker. The issue appears to be that the polyfill overrides the `self` variable, causing `ThreadWebWorker` to stop listening for messages from the DOM.

PR #499, opened 2024-11-19
Roll `Window.setGlobal` mutations of `globalThis.self`

lemonmade:
Fixes https://github.com/Shopify/remote-dom/issues/488

This PR partially reverts the changes from https://github.com/Shopify/remote-dom/pull/470. Specifically, it removes the mutation of `globalThis.addEventListener` (and friends), and `globalThis.self` from the side effectful `@remote-dom/core/polyfill` import. Our examples show users to import this module before using the `self` or `addEventListener` globals inside a Web Worker sandbox, which causes the global `self/addEventListener` to no longer be accessible for use in setting up a `postMessage` connection to a parent thread.

The change in https://github.com/Shopify/remote-dom/pull/470 definitely introduces a more complete polyfill of the DOM environment — with my revert, some invariants that hold in the browser no longer hold in the polyfill, like `globalThis.window === globalThis`. However, I think we need to hold off on changing the side-effectful import until the next breaking change, at which point I would prefer to change that behavior completely, and force consumers to call an exported method to install the polyfill. I've added a new static method to the `Window` polyfill, `setGlobalThis()`, which recreates the behavior of https://github.com/Shopify/remote-dom/pull/470, but in the form I expect to make this change (requires the consumer to explicitly call a method). I included a small snippet in the changeset for how I expect the admin to need to change its code, if it adopts a version between now and the breaking change.


e2e/basic.e2e.ts
```diff
@@ -6,13 +6,13 @@ import {test, expect} from '@playwright/test';
...
   ['iframe', 'htm'],
-  // ['iframe', 'react'],
-  // ['worker', 'vanilla'],
-  // ['worker', 'preact'],
-  // ['worker', 'svelte'],
+  ['iframe', 'react'],
+  ['worker', 'vanilla'],
+  ['worker', 'preact'],
+  ['worker', 'svelte'],
   // ['worker', 'vue'],
```
lemonmade, line 13: This example is still broken, but for unrelated reasons. I will fix it in a follow-up PR.

packages/core/source/polyfill.ts
```diff
@@ -72,4 +72,4 @@ hooks.removeAttribute = (element, name) => {
   updateRemoteElementAttribute(element, name);
 };
 
-export {hooks, window, type Hooks};
+export {hooks, window, Window, type Hooks};
```
lemonmade, line 75: I exported `Window` for convenience, so the patch for Admin can be a bit cleaner:

```js
import {window, Window} from '@remote-dom/core/polyfill';

Window.setGlobalThis(window);
```

vs

```js
import {window} from '@remote-dom/core/polyfill';
import {Window} from '@remote-dom/polyfill';

Window.setGlobalThis(window);
```

PR #511, opened 2024-11-25
Rd/legacy adapter

robin-drexler:


.changeset/beige-windows-appear.md
```diff
@@ -0,0 +1,9 @@
+---
+'@remote-dom/core': minor
+'@remote-dom/polyfill': minor
+'@remote-dom/preact': minor
+'@remote-dom/react': minor
+'@remote-dom/signals': minor
+---
+
+test
```
lemonmade, line 9: Make sure to update this before shipping

examples/remote-ui/README.md
```diff
@@ -0,0 +1,11 @@
+# Remote UI example
```
lemonmade, line 1: Did you consider just adding the remote UI example alongside the DOM-based remote environments in the other example? I am not sure if it would be nicer across the board, but it would be nice to avoid duplicating most of the host-side code between the two (and it helps reinforce that you can use Remote DOM exclusively on the host, while having remote-ui and Remote DOM running in the sandbox).

examples/remote-ui/app/index.html
```diff
@@ -0,0 +1,17 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" href="https://fav.farm/🛰️" />
+    <link rel="stylesheet" href="./style.css" />
+
+    <title>Getting started • Remote DOM</title>
```
lemonmade, line 8: If you are going to keep the examples separate, this should probably have a more specific title.

examples/remote-ui/app/remote/iframe.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" href="https://fav.farm/🛰️" />
+    <title>Getting started (remote) • Remote DOM</title>
```
lemonmade, line 6: Same as above re: a dedicated `title`, if this is going to be its own example.

examples/remote-ui/app/remote/remote.tsx
```diff
@@ -0,0 +1,24 @@
+/** @jsxRuntime automatic */
+/** @jsxImportSource react */
+import {createEndpoint, fromInsideIframe, retain} from '@remote-ui/rpc';
```
lemonmade, line 3: If you do merge this with the kitchen sink example, IMO it's fine to use the `@quilted/threads` library here instead of Remote UI’s RPC library.

examples/remote-ui/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "outDir": "build/typescript",
+    "jsx": "react-jsx",
+    "jsxImportSource": "preact"
+  },
+  "include": ["app"],
+  "references": [
+    {"path": "../../packages/core"},
+    {"path": "../../packages/preact"},
+    {"path": "../../packages/react"},
+    {"path": "../../packages/signals"}
```
lemonmade, line 13: Are these both needed?

package.json
```diff
@@ -17,19 +17,21 @@
     "version-bump": "changeset version && pnpm install --no-frozen-lockfile",
     "deploy": "pnpm publish -r",
     "example:getting-started": "pnpm run --filter example-getting-started start",
+    "example:remote-ui": "pnpm run --filter example-remote-ui start",
     "example:custom-element": "pnpm run --filter example-custom-element start",
     "example:kitchen-sink": "pnpm run --filter example-kitchen-sink start"
   },
   "devDependencies": {
+    "wait-on": "^8.0.1",
```
lemonmade, line 25: I didn’t notice this so far in my review — where is it used?

packages/core/README.md
```diff
@@ -751,3 +751,40 @@ This helper uses the following logic to determine whether a given property in th
 - If the property is an HTML element, it will be appended as a child in a slot named the same as the property (e.g., `<ui-button modal=${html`<ui-modal />`}>` becomes a `ui-modal` child with a `slot="modal"` attribute).
 - If the property starts with `on`, the value will be set as an event listener, with the event name being the lowercased version of the string following `on` (e.g., `onClick` sets a `click` event).
 - Otherwise, the property will be set as an attribute.
+
+### `@remote-dom/core/legacy`
+
+The `@remote-dom/core/html` package provides helpers for adapting between Remote DOM and [`remote-ui`, the previous version of this project](https://github.com/Shopify/remote-dom/discussions/267). These utilities are offered to help you transition to Remote DOM, while continuing to support existing code that expects `remote-ui`-style APIs.
```
lemonmade, line 757:
```suggestion
The `@remote-dom/core/legacy` package provides helpers for adapting between Remote DOM and [`remote-ui`, the previous version of this project](https://github.com/Shopify/remote-dom/discussions/267). These utilities are offered to help you transition to Remote DOM, while continuing to support existing code that expects `remote-ui`-style APIs.
```

packages/core/package.json
```diff
@@ -81,6 +91,7 @@
   },
   "dependencies": {
     "@remote-dom/polyfill": "workspace:^1.4.2",
+    "@remote-ui/core": "^2.0.0",
```
lemonmade, line 94: What do you think about an optional peer dependency for this instead? You'd need to also add it as a dev dependency (so types can be checked), but I think it might send the wrong message to have the new library explicitly depend on the old one like this, even if it’s for a good reason.

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+  type RemoteFragmentSerialization as LegacyRemoteFragmentSerialization,
+} from '@remote-ui/core';
+
+import type {
+  RemoteMutationRecord,
+  RemoteTextSerialization,
+  RemoteElementSerialization,
+  RemoteConnection,
+  RemoteNodeSerialization,
+} from '../types';
```
lemonmade, line 22:
```suggestion
} from '../types.ts';
```

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+ * sendChannelToRemoteEnvironment(channel);
+ * ```
+ */
+export function adaptToLegacyRemoteChannel(
+  connection: RemoteConnection,
+  options?: LegacyRemoteChannelOptions,
+): LegacyRemoteChannel {
+  const tree: Record<string, Array<{id: string; slot?: string}>> = {};
+
+  function persistNode(
```
lemonmade, line 64: Nit: can you move the returned function first, and have these functions appear after it? They all get hoisted up, but the reversed order puts the more important thing first, so you can see how these are used before seeing their implementation.

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+ *
+ * // Do something with the channel
+ * sendChannelToRemoteEnvironment(channel);
+ * ```
+ */
+export function adaptToLegacyRemoteChannel(
+  connection: RemoteConnection,
+  options?: LegacyRemoteChannelOptions,
+): LegacyRemoteChannel {
+  const tree: Record<string, Array<{id: string; slot?: string}>> = {};
```
lemonmade, line 62:
```suggestion
  const tree: Record<string, {id: string; slot?: string}[]> = {};
```

Also, since you are using this as a map, I would either flip it to a `Map`, or an object with a null prototype, to avoid the appearance of the "`tree.foo` might be defined on the prototype" footgun.

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+        const [id, props] =
+          payload as LegacyActionArgumentMap[typeof LEGACY_ACTION_UPDATE_PROPS];
+
+        const records = [];
+
+        for (const [key, value] of Object.entries(props)) {
+          if (isFragment(value)) {
+            const index = tree[id]?.findIndex(({slot}) => slot === key) ?? -1;
+
+            if (index !== -1) {
```
lemonmade, line 193: I'd probably flip this to `index >= 0`, so readers don’t need to handle double negation once they get down to the `else`

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+              records.push([
```
lemonmade, line 194: Should we check if the new fragment is actually different from the old one, before removing it?

packages/core/source/legacy/host.ts
lemonmade: Oops, I see there is no `else` block for this case, but I still think it's generally nicer to check for truthy conditions in cases like this.

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+
+            records.push([
+              MUTATION_TYPE_INSERT_CHILD,
+              id,
+              adaptLegacyFragmentSerialization(key, value, options),
+              tree[id]?.length ?? 0,
+            ] satisfies RemoteMutationRecord);
+          } else {
+            const index = tree[id]?.findIndex(({slot}) => slot === key) ?? -1;
+            if (index !== -1) {
```
lemonmade, line 209: Same comment as above

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+            }
+
+            records.push([
+              MUTATION_TYPE_INSERT_CHILD,
+              id,
+              adaptLegacyFragmentSerialization(key, value, options),
+              tree[id]?.length ?? 0,
+            ] satisfies RemoteMutationRecord);
+          } else {
+            const index = tree[id]?.findIndex(({slot}) => slot === key) ?? -1;
```
lemonmade, line 208: Should we lift this out since we have to deal with it in all cases? This also makes me wonder if we would be better off with a map of slot => index, so that we wouldn't need to loop through every child on each property update.

packages/core/source/legacy/host.ts
```diff
@@ -0,0 +1,335 @@
...
+}
+
+function adaptLegacyFragmentSerialization(
+  slot: string,
+  fragment: LegacyRemoteFragmentSerialization,
+  options?: LegacyRemoteChannelOptions,
+): RemoteElementSerialization {
+  return {
+    id: fragment.id,
+    element: 'remote-fragment',
```
lemonmade, line 326: Can we make this element an option you pass in to the main function, like it is for the [`slotProps.wrapper` option in React/ Preact component wrappers](https://github.com/Shopify/remote-dom/tree/main/packages/preact#:~:text=name%20using%20the%20%60slotProps.wrapper%60%20option)?

lemonmade (changes requested):
Had a bunch of requests, but this is looking really great 👍

examples/remote-ui/README.md
robin-drexler: I briefly looked into it and seemed kinda complicated for reasons I don't recall anymore tho. 😅 

I figured it's easier to just have a second example. I also assumed that we'd probably remove the remote-ui adapter and example at some point in the not too distant future.

examples/remote-ui/app/remote/remote.tsx
robin-drexler: Right, but nobody uses `@quilted/threads` with remote-ui. The idea was to have the example as close to what's actually out there today as possible.

examples/remote-ui/tsconfig.json
robin-drexler: Doesn't look like it. 👍 

removed!

package.json
robin-drexler: Good catch. I used this before I noticed that playwright actually supports multiple dev servers natively. Removed it

packages/core/source/legacy/host.ts
igor10k: `findIndex` is one of the rare cases when I personally prefer to always compare to `-1` as `-1` here holds a special meaning of "not found". I don't mind changing that to `>= 0` though.

packages/core/source/legacy/host.ts
igor10k: If the map is reduced to slot => index then we can't do a clean up when some slot is removed as we can't know if some other slot was a nested child of the removed slot.

packages/core/source/legacy/host.ts
igor10k: Added a `slotWrapper` property to `options`. But now I don't think we need it. We already have `elements` on the options. We can probably use it for the wrapper as well. Wdyt?

igor10k: good point

packages/core/package.json
igor10k: After a round of discussion we decided to extract it into a separate package. I went with `@remote-dom/compat` name.

robin-drexler:
@simontaisne another review from @lemonmade would be good, I think.

kumar303 (approved):
I'm lacking some context on `remote-dom` internals but it looks good to me. I played with the kitchen sink example and added some fragments to it in various places. Everything was working well ✨

packages/compat/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@remote-dom/compat",
+  "type": "module",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public",
+    "@remote-dom/registry": "https://registry.npmjs.org"
+  },
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
igor10k: I'm not familiar with how versioning works in this repo. My goal was to have a `1.0.0` release for the `compat` package. Should I put `0.1` here as the version and let changesets bump it to `1.0.0`?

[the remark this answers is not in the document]
lemonmade: Yes, I think you should do that. IIRC, as soon as you merge to `main` it will deploy the `0.1.0` version, and then changesets will create an additional PR that will bump it to `1.0.0` and manage the changelogs.

PR #521, opened 2024-12-19
Fix potentially wrong order of RemoteMutationRecords

### LINKED ISSUE #519: Insert- and AddMutations applied in wrong order at Receiver-side
### Brower
 (Chrome 131.0.6778.140)

### Error description

In some situation where an observed remote-mutation includes [interleaved insertions and additions](https://github.com/Shopify/remote-dom/blob/main/packages/core/source/elements/RemoteMutationObserver.ts#L43) of nodes, the (index-based) [application of these mutations](https://github.com/Shopify/remote-dom/blob/main/packages/core/source/receivers/RemoteReceiver.ts#L165) at receivers-side leads to accessing the children-collection at wrong indexes with the following downstream errors:

- insertion at wrong index

  ```
  TypeError: Cannot read properties of undefined (reading 'eventListeners')
      at Object.updateProperty (RemoteReceiver.mjs:125:1)
      at mutate (connection.mjs:25:1)
  ```

- removal at wrong index

  ```
  TypeError: Cannot read properties of undefined (reading 'id')
      at detach (RemoteReceiver.mjs:226:1)
      at Object.removeChild (RemoteReceiver.mjs:111:1)
      at mutate (connection.mjs:25:1)
  ```

I investigated on this issue and it comes up that this only happens if the amount of children is the same after the mutation. When this happens, the mutation observer seems to "batch" insertions and additions – first additions, then removals. This totally messes up the application of mutations at the receiver side, because it assumes an "ordered" mutations array.

### Example

```jsx
// From
<remote-parent>
  <remote-child key="1" />
  <remote-child key="2" />
</remote-parent>
```

```jsx
// To
<remote-parent>
  <remote-child key="2" />
  <remote-child key="3" />
</remote-parent>
```

The mutations collected by the `RemoteMutationObserver` are:

```js
[
  // remove child at index 2
  [MUTATION_TYPE_REMOVE_CHILD, 'parent-id', 2],
  // add child at index 0
  [MUTATION_TYPE_INSERT_CHILD, 'parent-id', { */ serialized child with key 3 */ }, 0],
]
```

When the `RemoteReceiver` tries to apply the first mutation, it wants to remove the child at index `2`, which does not exist 💥. Instead it should have done the insertion before **and then** the removal. In case of more than two mutations (which can be interleaved as well) all the removals are recorded before all insertions.

I don't know why the Mutation Observer API behaves this way (maybe some internal optimization), but this leads to unexpected (or even errors) on the host side.

### Possible solution

The `RemoteMutationObserver` could sort the mutations by their index [before sending](https://github.com/Shopify/remote-dom/blob/main/packages/core/source/elements/RemoteMutationObserver.ts#L103) them to the receiver. This should only be necessary if there are insertions and removals in one mutation and and amount of removals and insertions is equal.

mfal:
Resolves #519

### DIFF
--- a/.changeset/kind-sloths-hope.md
+++ b/.changeset/kind-sloths-hope.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/core': patch
+---
+
+fix: Fix errors when lists of remote elements change
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@remote-dom/core",
+  "name": "@mfalkenberg/remote-dom-core",
   "type": "module",
   "license": "MIT",
   "description": "A collection of DOM-based utilities for synchronizing elements between JavaScript environments",
@@ -7,7 +7,7 @@
     "access": "public",
     "@remote-dom/registry": "https://registry.npmjs.org"
   },
-  "version": "1.5.2",
+  "version": "1.5.6",
   "engines": {
     "node": ">=14.0.0"
   },
@@ -70,6 +70,7 @@
       ]
     }
   },
+  "files": ["build", "source"],
   "sideEffects": [
     "./source/polyfill.ts",
     "./build/cjs/polyfill.cjs",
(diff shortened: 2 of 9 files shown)


mfal:
This PR does not fix the underlying problem. See #548

PR #522, opened 2025-01-02
Does nothing when child in removeChild is not found

thomas-marcucci:


kumar303 (approved):
Seems like the right thing to do 👍 I believe Thomas has a shop to reproduce the error in a Shopify checkout and was able to see it fixed with this patch but I'll let him confirm.

thomas-marcucci:
> I believe Thomas has a shop to reproduce the error in a Shopify checkout and was able to see it fixed with this patch but I'll let him confirm.

Yes! I can prove it's working.

lemonmade (commented):
Can you give me a bit more detail on how this can actually happen in practice? This change being on the host implies that the remote environment is sending the request — what causes it to send a request to remove an element that does not exist?

This solution does work, but it feels like it hides a deeper bug.

thomas-marcucci:
The real bug is that calling `root.removeChild(someNode)` when `someNode` doesn't exist (or isn't a child of `root`) will remove the last child of `root`. It accepts a `-1` index. I'm wary of fixing that in case some extensions expect that behaviour (even if it's a bug).

The way this comes up from calling `root.removeChild(someNode)` on an empty root. Something like doing removeChild multiple times in a row. I think what I've seen is an extension removing nodes in a general setup phase then again after a query finished. I have a very contrived example that can show the exception and it's basically just calling `root.removeChild(node)` three times in a row. It's an issue with the remote code, but it causes a type error cause `undefined` is passed to `detach`.

thomas-marcucci:
If I'm understanding, then we should be fixing the missing child issue - but it will only be fixed for new extensions. I think it'd just be something like [this in root](https://github.com/Shopify/remote-dom/blob/c89427fa6e4176ce94fdc92bf14c95937d30256a/packages/core/src/root.ts#L844-L850) 
```diff
function removeChild(
   const {strict} = rootInternals;
 
   return perform(container, rootInternals, {
-    remote: (channel) =>
-      channel(
-        ACTION_REMOVE_CHILD,
-        (container as any).id,
-        container.children.indexOf(child as any),
-      ),
+    remote: (channel) => {
+      const childIndex = container.children.indexOf(child as any);
+      if (childIndex === -1) {
+        return;
+      }
+      return channel(ACTION_REMOVE_CHILD, (container as any).id, childIndex);
+    },
     local: () => {
       removeNodeFromContainer(child, rootInternals);
```
If child isn't found, do nothing (or something if we want to call attention to it). I don't understand everything in remote-ui, so I'm not sure if the local function needs a fix as well.

The error happens frequently, so we need to handle it somehow. Would wrapping the call to `currentReceiver.receive()` in checkout in a try-catch and doing some checks and logging there be sufficient?

lemonmade:
@thomas-marcucci yes, that is roughly what I was imagining. I think you can move the check outside of the `perform()` function, since to your point, it does also need to be applied to that `local` function too (it also does not seem to check whether the child is actually present in the parent, I think).

I still think your change on the host might be appropriate, but I'd want to pair it with this change to the remote, and a new API that can inform the consumer when such an invalid message has been received. This gives the consumer the full complement of support — we make it hard to make the mistake in the first place, by updating the remote environment, and we give them the ability to degrade gracefully in the case where invalid messages are received on the host (e.g., because an obnoxious extension was manually sending invalid postmessages, rather than using `RemoteRoot#removeChild()`.

packages/core/src/tests/receiver.test.ts
```diff
@@ -14,4 +19,45 @@ describe('createRemoteReceiver()', () => {
       expect(receiver).toHaveProperty('state', 'mounted');
     });
   });
+
+  describe('removeChild', () => {
```
thomas-marcucci, line 23: Figured I may as well leave this test in, but I'm okay removing it too 🤷

PR #523, opened 2025-01-09
Add type-check to snapit builds

igor10k:
Current snapit builds are missing the types as `type-check` is never run.


robin-drexler (approved):

.github/workflows/snapit.yml
```diff
@@ -23,5 +23,5 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         with:
-          build_script: pnpm build
+          build_script: pnpm type-check || true && pnpm build
```
lemonmade, line 26: Why the `|| true` bit — should it not fail if `pnpm type-check` exits unsuccessfully?

lemonmade (approved):

PR #526, opened 2025-01-22
add flush to remote connection

robin-drexler:


packages/core/source/elements/connection.ts
```diff
@@ -25,20 +25,24 @@ export class BatchingRemoteConnection {
   }
 
   mutate(records: any[]) {
-    let queued = this.#queued;
+    this.#queued ??= [];
+    this.#queued.push(...records);
 
-    if (queued) {
-      queued.push(...records);
+    this.#batch(() => {
```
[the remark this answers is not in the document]
robin-drexler: Good call. Changed and added a test. 

I had initially changed it because the way it's in main it'd call batch with old values again after flushing because this.#queue would be undefined, but the closed queued wasn't.

packages/core/source/elements/connection.ts
```diff
@@ -25,21 +25,31 @@ export class BatchingRemoteConnection {
...
-    queued = [...records];
-    this.#queued = queued;
-
     this.#batch(() => {
-      this.#connection.mutate(queued);
-      this.#queued = undefined;
+      if (this.#queued) {
+        this.#connection.mutate(this.#queued);
+        this.#queued = undefined;
+      }
```
lemonmade, line 41: Nit: can this call `this.flush()`/ both of them call some shared private function, since it’s the same "flush and empty queue" logic?

lemonmade (approved):

packages/core/source/elements/connection.ts
robin-drexler: Spoke too soon. I broke tests. Let me 👀

robin-drexler: seems like one test was flaky. I changed how he wait for next task to also use the message channel. 

I ran the test locally over 500 times without it failing once. 

```sh
for ((i=1; i<=500; i++)); do pnpm exec vitest --no-watch connection.test.ts || { echo "Command failed on attempt $i"; break; }; echo "Attempt $i succeeded"; done
```

With the previous `setTimeout` version, I could get it to break by running it 25 times in the first and 35 times in the second attempt.

PR #531, opened 2025-02-12
[wip] return subscriber promises

robin-drexler:


packages/async-subscription/src/tests/create.test.ts
```diff
@@ -72,6 +72,26 @@ describe('createRemoteSubscribable()', () => {
 
     expect(value).toBe(newValue);
   });
+
+  it('allows to catch async errors in the subscriber callback', async () => {
```
robin-drexler, line 76: With out the changes in this PR this crashes jest

```
 RUNS   async-subscription  packages/async-subscription/src/tests/create.test.ts
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Error: test".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
error Command failed with exit code 1.
```

With the changes it's green.

packages/async-subscription/src/tests/create.test.ts
```diff
@@ -72,6 +72,26 @@ describe('createRemoteSubscribable()', () => {
...
+    expect.assertions(1);
```
lemonmade, line 77: I find tests that rely on this really hard to read, since the assertion comes in the middle vs at the end. Is there any way we could adjust this — maybe by creating a promise that resolves with the `subscriber` callback — so that we can move the assertion to the end?

.changeset/violet-parents-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@remote-ui/async-subscription': patch
+---
+
+return subscriber result
```
lemonmade, line 5: Can you add a bit more detail about how a consumer might use this? You can use a simplified version of what you will do with it in our production apps.

lemonmade (approved):

PR #543, opened 2025-03-03
Try fixing adapter

oluwatimio:


packages/compat/source/adapter/host.ts
```diff
@@ -251,11 +251,23 @@ function adaptLegacyNodeSerialization(
...
   if (child.kind === LEGACY_KIND_TEXT) {
     return adaptLegacyTextSerialization(child);
+  } else if (child.kind === LEGACY_KIND_FRAGMENT) {
+    return adaptUndefinedLegacyFragmentSerialization(child);
   } else {
     return adaptLegacyComponentSerialization(child, options);
   }
 }
 
+function adaptUndefinedLegacyFragmentSerialization(props: any) {
```
oluwatimio, line 261: Remote-ui does the same 

Screenshot 2025-03-03 at 3 51 12 PM

Remote-dom with this cha

Screenshot 2025-03-03 at 3 55 19 PM
nge

PR #545, opened 2025-03-04
Handle fragments in adapter

igor10k:
The `remote-ui` to `remote-dom` adapter currently does not expect fragments to be passed and throws an exception when one is encountered. `remote-ui` host does accept fragments and maps them to an `undefined` component resulting in them not getting rendered. The adapter should mimic this behavior.

Note: Instead of an undefined component I went with mapping to a comment node. That achieves the same goal but also makes a bit more sense to me.

### DIFF
--- a/packages/compat/source/adapter/host.ts
+++ b/packages/compat/source/adapter/host.ts
@@ -1,5 +1,6 @@
 import {
   KIND_TEXT as LEGACY_KIND_TEXT,
+  KIND_COMPONENT as LEGACY_KIND_COMPONENT,
   KIND_FRAGMENT as LEGACY_KIND_FRAGMENT,
   ACTION_MOUNT as LEGACY_ACTION_MOUNT,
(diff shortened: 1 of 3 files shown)


packages/compat/source/adapter/host.ts
```diff
@@ -246,13 +249,26 @@ export function adaptToLegacyRemoteChannel(
...
+  switch (child.kind) {
+    case LEGACY_KIND_TEXT:
+      return adaptLegacyTextSerialization(child);
+    case LEGACY_KIND_COMPONENT:
+      return adaptLegacyComponentSerialization(child, options);
+    default:
+      return {
+        id: child.id,
+        type: NODE_TYPE_COMMENT,
+        data: '',
```
lemonmade, line 270:
```suggestion
        data: 'added by remote-ui legacy adaptor to replace a fragment rendered as a child',
```

lemonmade (approved):

PR #547, opened 2025-03-05
Handle reordering in adapter

igor10k:
The `remote-ui` to `remote-dom` adapter currently can't handle reordering which is achieved through `LEGACY_ACTION_INSERT_CHILD` instructions containing children with the same `id` that already exist in the DOM.

This PR mimics the [original logic](https://github.com/Shopify/remote-dom/blob/remote-ui/packages/core/src/receiver.ts#L169-L173), without keeping the `removed` child though as the adapter doesn't have access to the attached children.

#### Before

https://github.com/user-attachments/assets/b0ce4d5f-e93b-43cd-8e5b-a4133dfae163

#### After

https://github.com/user-attachments/assets/5b15aa9e-f8f5-4226-8d7b-ef4ff52d753d

### DIFF
--- a/.changeset/famous-olives-sit.md
+++ b/.changeset/famous-olives-sit.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/compat': patch
+---
+
+Handle reordering in adapter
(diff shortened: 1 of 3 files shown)


robin-drexler:
>without keeping the removed child 

does that have any implication?

igor10k:
> does that have any implication?

It seems to me that keeping the removed child was an optimization. Hoping to hear soon from @lemonmade

PR #556, opened 2025-03-20
Abstract global usage and export patch for window

henrytao-me:
Resolves https://github.com/Shopify/app-ui/issues/2145


packages/core/source/elements/RemoteElement.ts
```diff
@@ -268,7 +268,7 @@ export abstract class RemoteElement<
   Methods extends Record<string, (...args: any[]) => any> = {},
   Slots extends Record<string, any> = {},
   EventListeners extends Record<string, any> = {},
-> extends HTMLElement {
+> extends globalThis.HTMLElement {
```
henrytao-me, line 271: As long as we have a local scope for globalThis in realm, this should use the correct one.

PR #559, opened 2025-04-01
Issues with add then remove

lemonmade:


packages/signals/source/SignalRemoteReceiver.ts
```diff
@@ -319,8 +322,9 @@ export class SignalRemoteReceiver {
       attached.delete(child.id);
       parents.delete(child.id);
 
-      if (release && 'properties' in child) {
-        release(child.properties.peek());
+      if (release) {
+        if ('properties' in child) release(child.properties.peek());
+        if ('eventListeners' in child) release(child.eventListeners.peek());
```
lemonmade, line 327: The absence of this line is why this whole thing worked before — we were not releasing the function references attached as event listeners.

PR #575, opened 2025-06-11
Add early return in message listener for terminated endpoints

josemiguel-alvarez:
Terminated endpoints should not process new messages. Although this should not be happening as the listener is removed when terminating the endpoint, the early return will work as a safe guard to make sure we don't run into exceptions processing messages in terminated endpoints.


packages/rpc/src/endpoint.ts
```diff
@@ -247,7 +263,12 @@ export function createEndpoint<T>(
         const resolver = callIdsToResolver.get(callId);
 
         if (resolver == null) {
-          throw new MissingResolverError({callId, error, result});
+          throw new MissingResolverError({
+            callId,
+            error,
+            result,
+            groupingHash: 'MissingResolverError::FunctionResult',
```
lemonmade, line 270: I was chatting a bit about this with someone in relation Checkout's own error reporting: IMO, trying to distinguish two errors with a "grouping hash" or some other identifier passed in is a very weird pattern. If you need to differentiate between two errors, it should be two different error classes instead.

lemonmade (approved):
LGTM, although left a request to use separate error classes instead of a string identifier on the error, for that part of the change.

packages/rpc/src/endpoint.ts
josemiguel-alvarez: As we don't need to differentiate between two errors, I've updated MissingResolverError to always have the same groupingHash.

I'm happy to update if you think we should differentiate them.

PR #579, opened 2025-06-19
Adds previouslyResolved and terminated to MissingResolverError

thomas-marcucci:
Adds `previouslyResolved` and `terminated` to the `MissingResolverError` to help understand the context `MissingResolverError` is occurring under.

### Note
I'm unsure if we should clear `previouslyResolvedCallIds` only on terminate or if there are other circumstances it should be cleared. It's possible the use the same `callId` for a resolver multiple times successfully since it's removed from `callIdsToResolver` after it's used. So _may_ be a chance the error throws with `previouslyResolved = true` when that wasn't the cause 🤔

### DIFF
--- a/packages/rpc/src/endpoint.ts
+++ b/packages/rpc/src/endpoint.ts
@@ -49,9 +49,17 @@ export class MissingResolverError extends Error {
   readonly error?: Error;
   readonly result?: unknown;
   readonly groupingHash: string = 'RemoteUI::MissingResolverError';
+  readonly previouslyResolved: boolean = false;
+  readonly terminated: boolean = false;
 
-  constructor(message: {callId: string; error?: Error; result?: unknown}) {
-    const {callId, error, result} = message;
+  constructor(message: {
+    callId: string;
+    error?: Error;
+    result?: unknown;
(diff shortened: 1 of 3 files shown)


thomas-marcucci:
I just realized I mess something up with the test, but the rest of the PR is good to review 🙃

packages/rpc/src/endpoint.ts
```diff
@@ -239,11 +250,14 @@ export function createEndpoint<T>(
             callId,
             error,
             result,
+            previouslyResolved: previouslyResolvedCallIds.has(callId),
+            terminated,
```
josemiguel-alvarez, line 254: `terminated` will always be false, we check it at the beginning of this function.

thomas-marcucci: I added `terminated` _because_ it should always be `false` - just in case somehow it isn't.

lemonmade: IMO, this should be removed — the value can't change, and you aren't missing any possible edge cases where it could change. This is not the source of the issue.

lemonmade (changes requested):

PR #589, opened 2025-08-12
transferables for encoder

cpeddecord:
implementing a subset of widely available transferable objects ([MDN Link](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects#supported_objects)) for RPC's basic encoder.


lemonmade:
I would not personally want this to be the default behavior, since it is a destructive operation that the host has no way to control. In my more modern RPC library, I added a dedicated library function that lets a consumer mark transferable properties: https://github.com/lemonmade/quilt/blob/main/packages/threads/source/transfer.ts#L9 (though I would use a `WeakSet` to keep track of these, if I were doing it again today).

PR #590, opened 2025-08-19
fix error events not working

robin-drexler:
In polyfilled environments that use realms, it's currently not possible to listen to global errors and unhandled rejections. 

## `onerror` / `onunhandledrejection` handlers

e.g.
```js
window.onerror = (...args) => {
  console.log('### ERROR', args)
};

globalThis.onerror = (...args) => {
  console.log('### ERROR', args)
};
```

the following works btw. but is not exhaustive enough because libs like sentry and bugsnag attach themselves to `window` or `globalThis`
```js
onerror = (...args) => {
  console.log('### ERROR', args)
};
```

We now make this work by adding event listeners for those events the polyfill `window` instance that just forwards it to the `onerror` / `onunhandledrejection` functions. (thanks @igor10k for that idea!). Those events have been available for years so I think that's ok.

This also doesn't work at the moment inside extensions:

```js
globalThis.addEventListener("error", (...args) => {
  console.log('### ERROR', args);
});
```

This doesn't work because in realms the `globalThis.addEventListener` is *not* the one the worker is dispatching error events to. So we need to to dispatch those events manually from the worker to the realm global. 
However we can't dispatch the native event because the polyfill dispatchEvent implementation tries to override `target` on the event instance, which throws for native events.  So we need to also polyfill `ErrorEvent` and `PromiseRejectionEvent` which can be dispatched

### DIFF
--- a/.changeset/nasty-spoons-wonder.md
+++ b/.changeset/nasty-spoons-wonder.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/polyfill': patch
+---
+
+fix error events not working
--- a/packages/polyfill/source/ErrorEvent.ts
+++ b/packages/polyfill/source/ErrorEvent.ts
@@ -1,0 +1,20 @@
+import {Event} from './Event.ts';
+
+// https://html.spec.whatwg.org/multipage/webappapis.html#errorevent
+export class ErrorEvent extends Event {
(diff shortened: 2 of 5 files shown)


lemonmade (approved):
Great job!

PR #603, opened 2026-03-23
Add onMissingImplementationError hook to RemoteReceiverOptions

cpeddecord:
### tl;dr

when a node either mounts before or after a host is ready or has deleted a components remote method, we can sometimes be left with an invalid remote method call. This PR adds a hook for receivers to capture errors related to that error.

```
Node 2841 does not implement the hideOverlay() method
```

### Why

If we have isolated the error to faulty third-party code we may want to do something other than throw an error. This PR adds an optional hook that can accept the error or return a sentinel to fall back to the receiver's default thrown error.

### DIFF
--- a/packages/core/source/receivers/shared.ts
+++ b/packages/core/source/receivers/shared.ts
@@ -1,4 +1,29 @@
 /**
+ * Return this sentinel from error handlers to fall through to the default
+ * error case.
+ */
+export const THROW_DEFAULT: unique symbol = Symbol.for(
+  'remote-dom.throw-default',
+);
+
+/**
+ * Context passed to the `onMissingImplementationError` callback.
+ */
+export interface MissingImplementationContext {
+  /** The ID of the node the method was called on. */
+  id: string;
+  /** The name of the method that was called. */
+  method: string;
+  /** The arguments passed to the method call. */
+  args: readonly unknown[];
+  /**
+   * The element type of the node (e.g. `'s-modal'`, `'s-sheet'`), or
+   * `undefined` if the node is not an element or is not found in the tree.
+   */
+  element: string | undefined;
+}
+
+/**
  * Options that are useful for all remote receiver implementations. All of the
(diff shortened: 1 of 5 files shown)


packages/core/source/receivers/shared.ts
```diff
@@ -1,3 +1,28 @@
+/**
+ * Return this sentinel from error handlers to fall through to the default
+ * error case.
+ */
+export const THROW_DEFAULT: unique symbol = Symbol.for(
+  'remote-dom.throw-default',
+);
```
kumar303, line 7: Throwing is the default in this case but a default is a default, right?


```suggestion
export const DEFAULT: unique symbol = Symbol.for(
  'remote-dom.default',
);
```

packages/signals/source/SignalRemoteReceiver.ts
kumar303, line 1: This change doesn't seem to have any test coverage. I guess that's because there's no suite for this at all?

kumar303 (commented):
I'm struggling to imagine how this would be used. Is it just to observe and log the error? Is it so the host can render a no-op element? Is it to work around what feels like a bug in how elements are registered?

cpeddecord:
@kumar303 essentially that, yes. Right now a consumer would have to monkeypatch their own SignalRemoteReceiver implementation to inspect the the error and then duck type the emitted error to see if it matches.

packages/signals/source/SignalRemoteReceiver.ts
cpeddecord: I can backfill some tests here for sure.

kumar303:
@cpeddecord if it's observational only then why not just throw a more specific error that can be caught?

```ts
throw new NodeNotImplementedError(
  `Node ${id} does not implement the ${method}() method`,
  // Maybe some metadata if you need it?
  metadata: {...}
);
```

cpeddecord:
@kumar303, could do, that and all other errors.

PR #604, opened 2026-04-23
Flush BatchingRemoteConnection on a microtask by default

justinhenricks:
## Summary

Change the default `batch` function used by `BatchingRemoteConnection` to flush queued mutations on a **microtask** (via `queueMicrotask`, falling back to `Promise.resolve().then(...)`) instead of a macrotask via `MessageChannel`/`setTimeout`.

## Why

When the remote side runs logic like:

```ts
// Remote side (extension sandbox)
async perform() {
  fieldApi.setError('Invalid');   // queues a remote-dom mutation
}
```

...and the host runs logic like:

```ts
// Host side
await extension.perform();         // RPC awaits a response
focusFirstError(document);         // scans for [data-error] targets
```

...over an `@remote-ui/rpc`-style channel, the timeline under the old macrotask default is:

1. `setError` queues a remote-dom mutation (scheduled on a macrotask via `MessageChannel`)
2. RPC `await` resolves via the microtask queue
3. `focusFirstError` runs against stale DOM and finds 0 targets
4. The mutation arrives on the next macrotask, too late

Awaiting yields to the microtask queue but not the macrotask queue, so the RPC response always beats the mutation to the host.

In production, this is why checkout wouldn't auto-scroll to a buyer's invalid field when a blocking extension set an error: the auto-scroll ran before the error mutation landed, and there was nothing in the DOM to scroll to. We shipped the fix downstream in checkout-web ([shop/world#604053](https://github.com/shop/world/pull/604053), internal) by passing a custom `batch` option, but the macrotask default is surprising and easy to miss, so this patches it upstream.

Scheduling the flush on a microtask guarantees mutations land on the host before any `await`ed RPC response resolves.

## Changes

- `BatchingRemoteConnection`'s default `batch` now uses `queueMicrotask`, with a `Promise.resolve().then(...)` fallback.
- Updated the README section describing batching behavior, and noted how to restore macrotask batching via a custom `batch` option.
- Updated / added tests:
  - Renamed `waitForNextTask` → `waitForNextMicrotask`.
  - New test verifying flush happens before an `await`ed promise resolves.
  - New test verifying `queueMicrotask` is used.
  - New test verifying the `Promise.resolve().then(...)` fallback when `queueMicrotask` is unavailable.
- Added a `minor` changeset.

## Backwards compatibility

This is a behavioral change to the default, hence the **minor** bump. Consumers who depended on macrotask timing can restore it with a custom `batch` option (documented in both the changeset and the README).

### DIFF
--- a/.changeset/microtask-default-batching.md
+++ b/.changeset/microtask-default-batching.md
@@ -1,0 +1,21 @@
+---
+'@remote-dom/core': minor
+---
+
+Flush `BatchingRemoteConnection` mutations on a microtask by default
+
+The default `batch` function used by `BatchingRemoteConnection` now schedules flushes on a microtask (via `queueMicrotask`, falling back to `Promise.resolve().then(...)`) instead of a macrotask via `MessageChannel`/`setTimeout`.
+
+This ensures that DOM mutations made in the remote environment are delivered to the host before any `await`ed RPC response resolves, avoiding a class of bugs where the host sees an RPC result before the element updates that produced it (for example, a `perform()` callback setting an error on a field, only to have the host run its post-`await` logic before that error mutation arrives).
+
+If you were relying on the previous macrotask behavior, you can restore it by passing a custom `batch` option, e.g.:
+
+```ts
+new BatchingRemoteConnection(connection, {
+  batch: (flush) => {
+    const channel = new MessageChannel();
+    channel.port1.onmessage = () => flush();
+    channel.port2.postMessage(null);
+  },
+});
+```
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -474,7 +474,7 @@ root.connect(connection);
 
 The `RemoteConnection` object you receive from `RemoteReceiver.connection` is a simple object that immediately communicates all updates to the host environment. When using `RemoteMutationObserver`, documented above, this is not a major issue, since the `MutationObserver` API automatically batches DOM mutations. However, it can be more of a problem when using Remote DOM in a web worker (typically, with the `RemoteRootElement` wrapper), where no such batching is performed.
 
-To improve performance in these cases, you can use the `BatchingRemoteConnection` class, which batches updates from the remote environment that happen in the same JavaScript task. This class is a subclass of `RemoteConnection`, and can be used directly in place of the original connection object:
+To improve performance in these cases, you can use the `BatchingRemoteConnection` class, which batches updates from the remote environment. By default, queued mutations are flushed on a microtask (falling back to `Promise.resolve()`), so that mutations are observable on the host before any `await`ed RPC response resolves. You can override this by passing a custom `batch` function — for example, to defer flushes to a macrotask via `MessageChannel` or `setTimeout`. This class is a subclass of `RemoteConnection`, and can be used directly in place of the original connection object:
 
 ```ts
 import {
--- a/packages/core/source/elements/tests/connection.test.ts
+++ b/packages/core/source/elements/tests/connection.test.ts
@@ -1,11 +1,15 @@
 import '../../polyfill/polyfill.ts';
 
-import {describe, expect, it, vi, type MockedObject} from 'vitest';
+import {afterEach, describe, expect, it, vi, type MockedObject} from 'vitest';
 
 import {BatchingRemoteConnection, RemoteConnection} from '../../elements';
 
 describe('BatchingRemoteConnection', () => {
-  it('batches mutations', async () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('batches mutations on a microtask', async () => {
     const connection = createRemoteConnectionSpy();
     const batchingConnection = new BatchingRemoteConnection(connection);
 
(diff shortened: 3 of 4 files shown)


.changeset/microtask-default-batching.md
```diff
@@ -0,0 +1,21 @@
+---
+'@remote-dom/core': minor
```
robin-drexler, line 2: I almost wonder if that should be a breaking change. I'm leaning more towards no, but the thought has crossed my mind.

packages/core/source/elements/tests/connection.test.ts
```diff
@@ -23,34 +23,59 @@ describe('BatchingRemoteConnection', () => {
...
+    batchingConnection.mutate([1, 2, 3]);
 
-    vi.restoreAllMocks();
-    vi.useRealTimers();
+    expect(queueMicrotaskSpy).toHaveBeenCalledTimes(1);
+
+    queueMicrotaskSpy.mockRestore();
+  });
+
+  it('falls back to Promise.resolve().then() when queueMicrotask is unavailable', async () => {
```
robin-drexler, line 60: have you tested that the deletion works? Like if you take out the promise fallback does this test fail as expected?

robin-drexler (approved):
would really love to get @developit or @henrytao-me's 👀 on this

.changeset/microtask-default-batching.md
kumar303: It feels more like a bug fix to me

packages/core/source/elements/tests/connection.test.ts
```diff
@@ -23,34 +23,59 @@ describe('BatchingRemoteConnection', () => {
...
+    const connection = createRemoteConnectionSpy();
+    const batchingConnection = new BatchingRemoteConnection(connection);
+
+    batchingConnection.mutate([1, 2, 3]);
 
-    vi.restoreAllMocks();
-    vi.useRealTimers();
+    expect(queueMicrotaskSpy).toHaveBeenCalledTimes(1);
+
+    queueMicrotaskSpy.mockRestore();
```
kumar303: This needs to go in `afterEach` or try/finally.

packages/core/README.md
```diff
@@ -474,7 +474,7 @@ root.connect(connection);
...
 
-To improve performance in these cases, you can use the `BatchingRemoteConnection` class, which batches updates from the remote environment that happen in the same JavaScript task. This class is a subclass of `RemoteConnection`, and can be used directly in place of the original connection object:
+To improve performance in these cases, you can use the `BatchingRemoteConnection` class, which batches updates from the remote environment. By default, queued mutations are flushed on a microtask (via `queueMicrotask`, falling back to `Promise.resolve().then(...)`), so that mutations are observable on the host before any `await`ed RPC response resolves. You can override this by passing a custom `batch` function — for example, to defer flushes to a macrotask via `MessageChannel` or `setTimeout`. This class is a subclass of `RemoteConnection`, and can be used directly in place of the original connection object:
```
kumar303: nit to simplify:


```suggestion
To improve performance in these cases, you can use the `BatchingRemoteConnection` class, which batches updates from the remote environment. By default, queued mutations are flushed on a microtask (falling back to `Promise.resolve()`), so that mutations are observable on the host before any `await`ed RPC response resolves. You can override this by passing a custom `batch` function — for example, to defer flushes to a macrotask via `MessageChannel` or `setTimeout`. This class is a subclass of `RemoteConnection`, and can be used directly in place of the original connection object:
```

kumar303 (commented):
Minor request: In the PR description, I suggest adding a concrete example of how focusing the DOM after an error won't work since the DOM is stale. This example will concisely demonstrate the bug you're fixing.

An AI review also caught a minor issue worth looking into.

Otherwise the fix looks good ✨ 

## AI review

_Reviewed with Claude Opus 4.7 via pi._

<details>
<summary><strong>Explicit per-test setup</strong></summary>

- Consider splitting _"flushes mutations"_ into three focused tests: (a) `.flush()` drains synchronously, (b) `.flush()` cancels the pending microtask (no double-fire), (c) a subsequent `mutate()` schedules a fresh batch. Scenario (b) currently has no dedicated coverage.

</details>

kumar303 (approved):
^ whoops, meant to approve

packages/core/source/elements/tests/connection.test.ts
justinhenricks: yes! fails with `AssertionError: expected "spy" to be called 1 times, but got 0 times`

.changeset/microtask-default-batching.md
justinhenricks: yeah, I think if something were to break with this change that would be unexpected

packages/core/source/elements/tests/connection.test.ts
justinhenricks: good call

packages/core/README.md
justinhenricks: TY!

justinhenricks:
Thanks @kumar303 updated the PR description and addressed the feedback, as well as the AI note on splitting up the tests.

PR #607, opened 2026-04-27
prep snapit for OIDC

robin-drexler:
see https://github.com/Shopify/ui-extensions/blob/2026-07-rc/.github/workflows/deploy.yml#L26 as an example for a config that has been updated alredy

We're bumping snapit so it works with OIDC and we're adding the necessary permissions. 

We can't test this yet unfortunately. So we might have to make some adjustments soon.

### DIFF
--- a/.github/workflows/snapit.yml
+++ b/.github/workflows/snapit.yml
@@ -9,6 +9,10 @@ jobs:
   snapit:
     name: Snapit
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      id-token: write # Required for OIDC authentication
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/snapit.yml
```diff
@@ -9,19 +9,26 @@ jobs:
...
+      pull-requests: write
+      id-token: write # Required for OIDC authentication
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           token: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
 
       - uses: ./.github/workflows/actions/prepare
 
+      - name: Update npm to 11.7
```
robin-drexler, line 23: same as in ui-extensions. needed to ensure OIDC flows work

.github/workflows/snapit.yml
```diff
@@ -9,19 +9,26 @@ jobs:
...
+
       - name: Create snapshot
-        uses: Shopify/snapit@v0.0.14
+        uses: Shopify/snapit@v0.1.0
         env:
           GITHUB_TOKEN: ${{ secrets.SHOPIFY_GH_ACCESS_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_TOKEN: '' # Empty string forces OIDC
+          NPM_CONFIG_PROVENANCE: true
```
robin-drexler, line 31: https://github.com/Shopify/snapit#:~:text=Set%20NPM_CONFIG_PROVENANCE%3A%20true%20environment%20variable

robin-drexler:
closed in favor of https://github.com/Shopify/remote-dom/pull/608

PR #608, opened 2026-04-28
more prep for oidc

robin-drexler:
Npm only allows one singular action be be allow-listed so we need to move everything that publishes into `deploy.yml`

That means that each job now needs to check if it actually should be running by adding a bunch of checks in the `if` block.

this is a similar setup to what we have in ui-extensions: https://github.com/Shopify/ui-extensions/blob/2026-07-rc/.github/workflows/deploy.yml

See https://shopify.slack.com/archives/C01MXHNTT4Z/p1777390651870159?thread_ts=1777312754.951729&cid=C01MXHNTT4Z for more context

## 🎩 
I'm afraid we'll have to merge and try this. 🙃


.github/workflows/deploy.yml
```diff
@@ -2,19 +2,28 @@ name: Deploy
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - changeset-release/main
```
robin-drexler, line 7: needed because this is where preview-versions ran before

.nvmrc
```diff
@@ -1 +1 @@
-v20.11.0
+v20.20.0
```
robin-drexler, line 1: seems needed for npm in the version that we want

andrewmcgov (approved):
Looks good! Thanks for doing this.

PR #617, opened 2026-07-15
Add Web Platform Test runner infrastructure

andrewiggins:
## Summary

- add a private `@remote-dom/wpt-runner` workspace package for running selected upstream `testharness.js` HTML files against `@remote-dom/polyfill` without modifying the upstream files
- pin and checksum the downloaded WPT archive, validate it before extraction, and use revision-addressed caches that are safe to share across concurrent worktrees
- add a canonical TSV capability inventory that distinguishes supported cases from documented deferred cases and rejects missing, duplicate, or unlisted results
- isolate each test file in a fresh worker with a new polyfill `Window`, private result transport, and worker-specific network restrictions
- add Playwright CLI and debug-page entry points plus a dedicated CI job with cached WPT sources

## Architecture

The browser control page parses each selected WPT file and turns supported top-level markup and classic scripts into generated source in their original order. The CLI reuses one control page across files, but each file runs in a fresh module worker with a new `Window` from `@remote-dom/polyfill`.

Runner results travel over a private `MessagePort` rather than the worker's public message channel. A worker-specific Content Security Policy blocks network connections and nested workers while allowing the same-origin module loading and generated-code execution required by the runner. Served WPT files use ETag revalidation to avoid repeatedly transferring shared resources such as `testharness.js`.

This infrastructure tests the worker-side DOM implementation directly. It intentionally does not construct a Remote DOM host receiver or claim transport and rendering behavior. The initial capability table contains only the runner smoke fixture. Follow-up PRs can add upstream DOM, HTML, and SVG files one capability at a time.

## Validation

- `pnpm run lint`
- `pnpm run type-check`
- `pnpm exec vitest run`
- `pnpm run test:wpt`
- verified a cold archive download and extraction, including the revision-root completion-marker layout
- verified concurrent preparation shares one validated archive and recovers abandoned locks
- verified `WPT_ROOT=/path/to/wpt` reuses a local checkout without downloading
- verified worker networking and nested workers are blocked
- verified the CLI reuses its control page while creating a fresh worker for each file

## Changeset

An empty changeset is included because `@remote-dom/wpt-runner` is private test infrastructure and does not release a public package.


andrewiggins:
Superseded by the three-PR replacement stack:

1. Source preparation: https://github.com/Shopify/remote-dom/pull/630
2. Interactive runner: https://github.com/Shopify/remote-dom/pull/631
3. Capability enforcement and CI: https://github.com/Shopify/remote-dom/pull/632

The replacement stack gives each PR a self-contained executable outcome. This PR is closed without merging.

PR #621, opened 2026-07-16
Polyfill remote-dom FormData using the native Worker implementation

developit:
## Summary
Bridge remote-dom forms to the native `FormData` implementation available in Web Workers, instead of reimplementing the API. This keeps the polyfill small while preserving native behavior for storage, iteration, file handling, and multipart request interoperability.

## What changed
- Capture the environment’s native `FormData` constructor and use it to build form data from remote-dom controls.
- Populate `FormData` from the common subset of successful controls supported by remote-dom.
- Detect checkable controls by property presence so custom elements participate correctly.
- Avoid triggering unrelated property reads when controls are omitted from submission.

## Notes
- The implementation intentionally stays minimal and leans on the browser’s native `FormData` semantics.
- Custom checked controls are supported via property-based detection rather than input type heuristics.


packages/polyfill/source/FormData.ts
```diff
@@ -0,0 +1,39 @@
...
+      if (type === 'submit') continue;
+
+      if (
+        'checked' in control &&
+        !(control.checked ?? control.hasAttribute('checked'))
+      ) {
+        continue;
+      }
+
+      data.append(name, control.value ?? control.getAttribute('value') ?? '');
```
frehner, line 32: checked elements submit `"on"` as a fallback if the value attribute isn't there

> Otherwise, if the field element is an [input](https://html.spec.whatwg.org/multipage/input.html#the-input-element) element whose [type](https://html.spec.whatwg.org/multipage/input.html#attr-input-type) attribute is in the [Checkbox](https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox)) state or the [Radio Button](https://html.spec.whatwg.org/multipage/input.html#radio-button-state-(type=radio)) state:
>
>If the field element has a [value](https://html.spec.whatwg.org/multipage/input.html#attr-input-value) attribute specified, then let value be the value of that attribute; otherwise, let value be the string "on".

step 7.1 here https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set

PR #622, opened 2026-07-16
Fix insertBefore leaving the previous sibling pointing past the inserted node

airhorns:
## The bug

`ParentNode.insertInto(child, before)` relinks the new child's `NEXT`/`PREV` and the reference node's `PREV`, but only repoints the **previous sibling's `NEXT`** when the reference node is the first child:

```js
if (before) {
  child[NEXT] = before;
  child[PREV] = before[PREV];
  if (before[PREV] === null) this[CHILD] = child;  // only insert-at-front is handled
  before[PREV] = child;                            // missing: else child[PREV][NEXT] = child
}
```

Insert before any *later* child and the forward sibling chain skips the new node.

That chain is exactly what `querySelector` / `querySelectorAll` recurse through (`within[CHILD]`, then `[CHILD]` / `[NEXT]`), and what `firstChild` + `nextSibling` traversal walks. So an affected node becomes unreachable from every selector query and sibling walk — while `childNodes` (a spliced array), `parentNode`, `previousSibling`, `isConnected`, and the host mirror (built from the `insertChild(parent, child, index)` hook) all still report it as present.

The node renders correctly on the host and is invisible to the remote environment's own lookups, with **no error raised**. `replaceChild` (`removeChild` + `insertInto(new, next)`), `prepend` of more than one node, and `ChildNode.after` on a non-last child all route through the same branch and were affected too.

The divergence, on `main`, from a hand-rolled host mirror fed by the real `insertChild` hook:

| | sandbox chain | host mirror | `querySelector('s-modal')` |
|---|---|---|---|
| before | `[s-page, style]` ❌ | `[s-page, s-modal, style]` | `null` ❌ |
| after | `[s-page, s-modal, style]` ✅ | `[s-page, s-modal, style]` | `s-modal` ✅ |

### DIFF
--- a/.changeset/polyfill-insert-before-sibling-chain.md
+++ b/.changeset/polyfill-insert-before-sibling-chain.md
@@ -1,0 +1,11 @@
+---
+'@remote-dom/polyfill': patch
+---
+
+Fix `insertBefore()` leaving the previous sibling pointing past the inserted node
+
+`ParentNode.insertBefore()` relinked the new child's `NEXT`/`PREV` and the reference node's `PREV`, but only repointed the previous sibling's `NEXT` when the reference node was the first child. Inserting before any _later_ child therefore left the forward sibling chain skipping the new node.
+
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

andrewiggins:
Turns out I also had this fix in #620 so going to close this since I merged that one. However I didn't have quite as extensive of a test suite as you have here so I opened #694 to preserve it.

PR #624, opened 2026-07-28
Implement MutationObserver in the DOM polyfill

airhorns:
## What changed

`@remote-dom/polyfill` currently installs and documents a `MutationObserver` constructor, but the constructor is an empty stub: `new MutationObserver(callback).observe(node)` throws because `observe` is undefined. This is especially easy to miss when remote code is type-checked against `lib.dom`.

This implements the observer surface over the polyfilled tree:

- asynchronous, batched child-list, attribute, and character-data records
- `subtree`, `attributeFilter`, and old-value options
- `observe()`, `disconnect()`, and `takeRecords()`
- standard callback and observation-option validation

Mutation delivery is integrated alongside the existing remote synchronization hooks, so observation does not change what is sent to a Remote DOM host.

## Inactive-path performance

Mutation types maintain separate active-registration flags. Mutation sites check the corresponding flag before capturing old values or allocating a record, `Map`, or `NodeList`, and before walking ancestors. The registration `WeakMap` is created lazily and released again when the final observer disconnects. A child-list or character-data observer therefore does not add bookkeeping to attribute writes.

In a local Node 20.20 benchmark of one million repeated `setAttribute()` calls with no active observers (seven measured runs after warm-up), the median changed from 66.9ms to 47.0ms compared with the initial implementation in this PR, about 30% faster.

## Validation

- `pnpm exec vitest run` (180 tests)
- `pnpm lint`
- `pnpm type-check`
- `mise exec node@20.20.0 -- pnpm --filter @remote-dom/polyfill build`

### DIFF
--- a/.changeset/quiet-observers-watch.md
+++ b/.changeset/quiet-observers-watch.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/polyfill': minor
+---
+
+Implement the standard `MutationObserver` methods for child, attribute, and character-data changes in the polyfilled DOM.
--- a/packages/polyfill/source/Attr.ts
+++ b/packages/polyfill/source/Attr.ts
@@ -11,6 +11,7 @@ import {
 } from './constants.ts';
 import {Node} from './Node.ts';
 import type {Element} from './Element.ts';
+import {queueMutationRecord} from './MutationObserver.ts';
 
 export class Attr extends Node {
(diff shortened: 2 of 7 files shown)


andrewiggins:
Now that #652 has landed, I'd like to implement this as a built-in Window extension consumers can opt into

andrewiggins (approved):
Gonna go ahead and merge this and then build on top of it

PR #625, opened 2026-07-28
Add getElementsByClassName to the DOM polyfill

airhorns:
## What changed

Remote code is commonly type-checked against `lib.dom`, but `document.getElementsByClassName()` and `element.getElementsByClassName()` are missing at runtime from `@remote-dom/polyfill`. This adds both standard hosts (deliberately not `DocumentFragment`) and returns matches in tree order.

The lookup tokenizes ASCII whitespace, requires every requested class, removes duplicate search tokens, and compares class names literally. It does not build a CSS selector, so class names containing selector punctuation such as `.` and `:` work correctly.

The returned value follows the polyfill's existing concrete `NodeList` collection model. Live `HTMLCollection` behavior remains a separate compatibility concern, as it does in #620.

## Validation

- `pnpm exec vitest run` (179 tests)
- `pnpm lint`
- `pnpm type-check`
- `mise exec node@20.20.0 -- pnpm --filter @remote-dom/polyfill build`

### DIFF
--- a/.changeset/fuzzy-classes-find.md
+++ b/.changeset/fuzzy-classes-find.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/polyfill': minor
+---
+
+Add `getElementsByClassName()` to polyfilled documents and elements.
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):

andrewiggins (approved):
Thanks for the PR! Gonna follow up with some improvements, but this is a great start.

PR #627, opened 2026-07-28
Forward focus and scrollIntoView to the host

airhorns:
## What changed

`focus()` and `scrollIntoView()` are missing from polyfilled elements, even though Remote DOM already has a method-call channel and both operations can be performed by the rendered host element. Remote code type-checked against `lib.dom` therefore fails at runtime before it can use that channel.

This adds a general imperative-method hook to `@remote-dom/polyfill`, implements `focus()` and `scrollIntoView()` on the polyfill's shared `Element` abstraction, and wires the core polyfill hook to the connected `RemoteConnection`. Options are forwarded unchanged. Standalone or disconnected polyfilled elements keep browser-like no-op behavior when no host hook is available.

The core regression test connects a real `RemoteRootElement` and `RemoteReceiver`, registers the host implementation, and verifies both method calls end to end.

These methods live on the shared `Element` class because that is the current concrete class returned for ordinary and custom HTML elements on `main`. #626 separately proposes concrete HTML constructor identity.

`getBoundingClientRect()` is deliberately not included: it is synchronous and returns host layout data, which cannot be faithfully implemented over an asynchronous remote connection.

## Validation

- `pnpm exec vitest run` (178 tests)
- `pnpm lint`
- `pnpm type-check`
- `mise exec node@20.20.0 -- pnpm --filter @remote-dom/polyfill --filter @remote-dom/core build`

### DIFF
--- a/packages/polyfill/source/Element.ts
+++ b/packages/polyfill/source/Element.ts
@@ -1,6 +1,7 @@
 import {
   NS,
   ATTRIBUTES,
+  HOOKS,
   HTML_NAMESPACE,
   NODE_TYPE_ELEMENT,
   type NamespaceURI,
@@ -87,20 +88,12 @@ export class Element extends ParentNode {
     return sib;
   }
 
-  focus(options?: FocusOptions) {
-    if (options === undefined) {
-      this.callHostMethod('focus');
-    } else {
-      this.callHostMethod('focus', options);
-    }
+  focus(...args: [options?: FocusOptions]) {
+    this[HOOKS].callMethod?.(this as any, 'focus', ...args);
   }
 
-  scrollIntoView(arg?: boolean | ScrollIntoViewOptions) {
-    if (arg === undefined) {
-      this.callHostMethod('scrollIntoView');
-    } else {
-      this.callHostMethod('scrollIntoView', arg);
-    }
+  scrollIntoView(...args: [arg?: boolean | ScrollIntoViewOptions]) {
+    this[HOOKS].callMethod?.(this as any, 'scrollIntoView', ...args);
   }
 
   setAttribute(name: string, value: string) {
--- a/packages/polyfill/source/Element.ts
+++ b/packages/polyfill/source/Element.ts
@@ -87,6 +87,22 @@ export class Element extends ParentNode {
     return sib;
   }
 
+  focus(options?: FocusOptions) {
+    if (options === undefined) {
+      this.callHostMethod('focus');
+    } else {
+      this.callHostMethod('focus', options);
+    }
+  }
+
+  scrollIntoView(arg?: boolean | ScrollIntoViewOptions) {
+    if (arg === undefined) {
+      this.callHostMethod('scrollIntoView');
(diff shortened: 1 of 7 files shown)


JoviDeCroock (approved):

packages/polyfill/source/Element.ts
```diff
@@ -67,6 +67,22 @@ export class Element extends ParentNode {
...
+      this.callHostMethod('focus', options);
+    }
+  }
+
+  scrollIntoView(arg?: boolean | ScrollIntoViewOptions) {
+    if (arg === undefined) {
+      this.callHostMethod('scrollIntoView');
+    } else {
+      this.callHostMethod('scrollIntoView', arg);
+    }
```
developit:
```suggestion
    this[HOOKS].callMethod?.(this as any, 'scrollIntoView', arg);
```

packages/polyfill/source/Element.ts
```diff
@@ -67,6 +67,22 @@ export class Element extends ParentNode {
     return sib;
   }
 
+  focus(options?: FocusOptions) {
+    if (options === undefined) {
+      this.callHostMethod('focus');
+    } else {
+      this.callHostMethod('focus', options);
+    }
```
developit:
```suggestion
    this[HOOKS].callMethod?.(this as any, 'focus', arg);
```

packages/polyfill/source/Node.ts
```diff
@@ -36,6 +36,10 @@ export class Node extends EventTarget {
     return this[OWNER_DOCUMENT].defaultView[HOOKS];
   }
 
+  protected callHostMethod(method: string, ...args: unknown[]) {
+    this[HOOKS].callMethod?.(this as any, method, ...args);
+  }
+
```
developit: This makes the method callable from userland. Callsites must just use `this[HOOKS].callMethod?.(this as any, 'scrollIntoView', arg)` directly (see other suggestion comments).

```suggestion
```

developit (changes requested):

packages/polyfill/source/Node.ts
airhorns: Done in 726ae5f — `callHostMethod` is gone and both call sites invoke `this[HOOKS].callMethod?.(this as any, …)` directly. Also rebased onto main to pick up the erasable-syntax import changes.

packages/polyfill/source/Element.ts
airhorns: Applied, with one tweak: the argument is spread through a rest tuple (`focus(...args: [options?: FocusOptions])`) so a bare `element.focus()` reaches the host as `callMethod(el, 'focus')` rather than `callMethod(el, 'focus', undefined)` — keeps the arity the host sees identical to what the caller wrote.

packages/polyfill/source/Element.ts
airhorns: Same as above — inlined via a rest tuple to preserve arity.

PR #631, opened 2026-08-21
Add an interactive Web Platform Test runner

andrewiggins:
## Summary

This PR is 2 of 3 (#630, #632) that adds a Web Platform Test runner to this repo, so we can validate our polyfills spec compliance against that test suite.

This PR contains the actual runner that executes the WPT test (`main.ts`). It's a webpage (`index.html`) that creates a Web Worker (`run-worker.ts`), loads our DOM polyfill into that worker (`worker.ts`), and parses the WPT test into JS instructions (`adapter.ts`) that it executes in that worker with the DOM polyfill. It uses the WPT test harness to execute the test inside the worker and get the results.

Specifically, this PR:
- securely serve prepared WPT files with canonical path and symlink containment
- adapt supported top-level HTML/SVG markup and classic scripts into generated source
- execute each selected file in a fresh worker and polyfill `Window`
- keep results on a private transferred `MessagePort`
- make the first terminal event final and always clean up the worker
- block worker networking and nested workers with a worker-specific CSP
- provide an interactive page for running tests and inspecting source and results

## Stacked dependency

This PR is stacked on `wpt-runner-source-preparation`, which prepares and validates the pinned WPT checkout.

## Executable outcome

Run `pnpm wpt:dev`, enter a WPT path, and inspect the result, warnings, console output, original markup, harness source, and generated source.

This PR intentionally does not add capability policy, Playwright CLI automation, or WPT CI.

## Validation

- `pnpm run lint`
- `pnpm run type-check`
- `pnpm exec vitest run`
- interactive three-case smoke execution
- repeated-run fresh-worker validation

## Changeset

An empty changeset records that this private test infrastructure releases no public package.

### DIFF
--- a/packages/wpt-runner/fixtures/runner-smoke.html
+++ b/packages/wpt-runner/fixtures/runner-smoke.html
@@ -1,0 +1,24 @@
+<!doctype html>
+<title>Remote DOM WPT runner smoke test</title>
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<section>Static HTML</section>
+<svg viewBox="0 0 10 10">
+  <rect width="10" height="10" fill="green"></rect>
+</svg>
+<script>
+  test(() => {
+    assert_equals(document.body.children[0].localName, 'section');
+    assert_equals(document.body.children[0].textContent, 'Static HTML');
+    assert_equals(
+      document.body.children[1].namespaceURI,
+      'http://www.w3.org/2000/svg',
+    );
+    assert_equals(
+      document.body.children[1].children[0].getAttribute('fill'),
+      'green',
+    );
+    assert_equals(location.pathname, '/__runner__/runner-smoke.html');
+    assert_equals(location.search, '?runner=smoke');
+  }, 'runner smoke');
+</script>
--- a/packages/wpt-runner/fixtures/runner-smoke.html
+++ b/packages/wpt-runner/fixtures/runner-smoke.html
@@ -7,6 +7,46 @@
   <rect width="10" height="10" fill="green"></rect>
 </svg>
 <script>
+  promise_test(async (test) => {
+    await promise_rejects_js(
+      test,
+      TypeError,
+      fetch('/__wpt-file?path=__runner__/runner-smoke.html'),
+    );
+  }, 'runner worker CSP blocks fetch');
+
+  promise_test(
+    () =>
+      new Promise((resolve, reject) => {
+        let nestedWorker;
+        try {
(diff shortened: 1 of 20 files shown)


packages/wpt-runner/fixtures/runner-smoke.html
```diff
@@ -0,0 +1,64 @@
+<!doctype html>
```
andrewiggins, line 1: This file is just a test file we can use to smoke test the runner. It demonstrates what a classic WPT test looks like.

packages/wpt-runner/public/wpt-runner/runtime-shims.js
```diff
@@ -0,0 +1,40 @@
+const __WPT_SVG_NAMESPACE__ = 'http://www.w3.org/2000/svg';
```
andrewiggins, line 1: The WPT `testharness` needs some additional featues polyfill doesn't have yet so this adds the required shims so tests can run. Over time we might implement some of these in polyfill so this file might go away. Left that as out of scope for now.

packages/wpt-runner/src/adapter.ts
```diff
@@ -0,0 +1,318 @@
+const SVG_NAMESPACE = 'http://www.w3.org/2000/svg';
```
andrewiggins, line 1: This file parses a WPT HTML test file and converts it to JS instructions we can execute in a worker against our Remote DOM polyfill. So a declaration like `<svg viewBox>` becomes a `createElement('svg')` and `setAttribute(viewBox` instruction.

It also handles downloading the script tags a test might reference as some times write their tests in a separate JS file. We download an execute those directly in the order they are in the HTML file.

packages/wpt-runner/src/worker.ts
```diff
@@ -0,0 +1,168 @@
...
+    installWindowPostMessage(window);
+    installHarnessReadyHook();
+    installConsoleTransport();
+    respond({type: 'ready'});
+
+    const AsyncFunction = Object.getPrototypeOf(async function () {})
+      .constructor as new (source: string) => () => Promise<void>;
+    await new AsyncFunction(
+      `${request.source}\n//# sourceURL=wpt:${request.path}`,
+    )();
```
andrewiggins, line 59: Here's the meat. We take the generated bundle of `createElement` instructions that `buildWptBundle` in `adapter.ts` created, and execute them in an async function whose scope is this worker, upon which RemoteDom's polyfills are installed.

marvinhagemeister (approved):

PR #633, opened 2026-08-21
Modernize development and CI infrastructure

andrewiggins:
## Summary

Modernize Remote DOM’s repository infrastructure while preserving its existing package contracts and supported runtime behavior.

This PR broadly:

- Updates Node, pnpm, TypeScript, Playwright, jsdom, Rollup, Vitest, and related build tooling.
- Centralizes shared dependencies and adds pnpm supply-chain protections.
- Simplifies dependency installation and improves CI build and coverage validation.
- Groups routine Dependabot updates.
- Refreshes contributor and security documentation.
- Removes the obsolete Shipit release configuration.

## Additional context

TypeScript 7 required a few source-level compatibility adjustments:

- The core HTML normalization and Preact renderer changes only make their existing types explicit.
- TypeScript 7’s DOM types require `CustomElementRegistry.initialize()`. The polyfill now provides a no-op implementation matching its existing no-op `upgrade()` method. This does not add scoped custom-element registry support; it just makes TypeScript happy.

Prettier remains at 3.3.3 because upgrading it would introduce unrelated formatting churn.

No changeset is included because these changes are infrastructure-focused and do not intentionally change public package behavior.

## Validation

- Lint, type checking, and all package builds pass.
- All 174 Vitest tests and 13 Playwright tests pass.
- V8 coverage reporting and package line thresholds pass.
- Published package output was compared against `main`; remaining differences were reviewed as TypeScript compatibility or compiler-output changes.


marvinhagemeister (approved):

kumar303 (approved):
Thanks for working on this, Andre ✨ 

I also had an agent verify the package build. It matched the published versions on `registry.npmjs.org` so all good 👍 

The agent flagged a few things below that may be worth looking at. The examples were already pretty bit-rotted, I think 😓 

## 🤖 👯‍♀️ 

<details>
<summary>1. <code>pnpm example:getting-started</code> and <code>pnpm example:custom-element</code> are broken</summary>

`vite` was removed from the root `devDependencies`, but `examples/getting-started` and `examples/custom-element` both run `"start": "vite"` without declaring it. Only `examples/kitchen-sink` declares `vite` (`^5.4.0`), which is why Playwright E2E still passes (it serves kitchen-sink).

Confirmed:

```
$ pnpm --filter example-getting-started exec sh -c 'command -v vite || echo "VITE NOT FOUND"'
VITE NOT FOUND
$ pnpm --filter example-kitchen-sink exec sh -c 'command -v vite'
examples/kitchen-sink/node_modules/.bin/vite
```

**Fix:** add `vite` to the `devDependencies` of `examples/getting-started` and `examples/custom-element` (or restore it at the root).
</details>

<details>
<summary>2. <code>npm install -g npm@11.7</code> removal may break OIDC publishing</summary>

`deploy.yml` drops three `Update npm to 11.7` steps, but `NPM_TOKEN: ''` is still set to force OIDC trusted publishing — and that npm upgrade was almost certainly there because OIDC requires npm ≥ 11.5.1. The publish paths are `pnpm publish -r`, `pnpm changeset publish`, and `Shopify/snapit@v0.1.0` (third party, publishes on our behalf).

**Fix:** confirm pnpm 10.28 and snapit handle OIDC natively without a modern global npm, otherwise the release job fails at publish time.
</details>

<details>
<summary>3. <code>corepack enable</code> removal relies on <code>pnpm/action-setup</code> reading <code>packageManager</code></summary>

The prepare action no longer runs `corepack enable && pnpm -v`. `pnpm/action-setup@v3` (already the first step) does resolve the version from the root `packageManager` field, so this should work — but the version is no longer echoed anywhere, so a resolution mismatch would surface as a confusing lockfile/install error rather than an obvious one.

**Fix (optional):** keep a `pnpm -v` echo step for debuggability.
</details>

<details>
<summary>Bonus: verify <code>catalog:</code> resolution on the snapit publish path</summary>

Not strictly a repo-only issue — it is the one change here with catastrophic blast radius. `htm: "catalog:"` now sits in the published `dependencies` of `core`, `preact`, and `react`. `pnpm pack` resolves it correctly:

```json
"dependencies": {"@remote-dom/polyfill": "^1.5.1", "htm": "^3.1.1"}
```

But that only holds for pnpm-driven publishes. If `Shopify/snapit@v0.1.0` ever shells out to `npm publish`, a literal `"htm": "catalog:"` ships and every install fails with `Unsupported URL Type "catalog:"`.

**Fix:** run `/snapit` on this PR and inspect the published manifest before merging.
</details>

PR #652, opened 2026-08-31
Composable polyfill extensions

olavoasantos:
### TL;DR

Add a composable extension API to `@remote-dom/polyfill`.

`Window.with(...extensions)` creates a reusable `Window` subclass without changing the zero-argument `Window` constructor. Each extension is a function that runs once per window during construction and can return hooks that subscribe to DOM operations:

```ts
const ExtendedWindow = Window.with(
  mutationObserverExtension,
  cssomExtension,
);

const window = new ExtendedWindow();
```

Window API assignments are plain property writes, so later extensions can override APIs installed by earlier ones. Extension hooks are additive and run in installation order. The existing `window[HOOKS]` API is now documented as the legacy integration path; it remains supported and runs after extension hooks.

### What changed?

Extensions are plain functions:

```ts
type WindowExtension = (window: Window) => Partial<Hooks> | void;
```

The function has two distinct scopes:

- The function body runs once per `Window` instance, during construction. It receives the window and can install or replace DOM APIs on it, such as a `MutationObserver` implementation, while keeping state in a closure scoped to that window. If the window is later installed with `Window.setGlobal()`, those APIs become globals.
- The returned hooks run for the lifetime of the window. They subscribe to runtime operations like `createElement`, `setAttribute`, and `insertChild`, and are called each time the corresponding operation happens. Returning hooks is optional when an extension only needs construction-time setup.

These parts compose differently. Window API assignments are last-write-wins. Hooks are additive and dispatched per window in installation order, including across chained `.with()` calls.

The existing low-level API is unchanged:

```ts
window[HOOKS].createElement = (element) => {
  // ...
};
```

Legacy hooks are resolved at dispatch time and run after extension hooks, so existing integrations like `@remote-dom/core/polyfill` continue to work without changes. The polyfill README documents the extension API, its two scopes, these composition rules, and the legacy status of `window[HOOKS]`.

### Why make this change?

The existing `window[HOOKS]` API gives each DOM operation a single mutable callback. Composing integrations requires every consumer to preserve and invoke the previous callback:

```ts
const previous = window[HOOKS].createElement;

window[HOOKS].createElement = (element, namespace) => {
  previous?.(element, namespace);

  // Handle the operation.
};
```

This is powerful, but fragile:

- forgetting to preserve the previous hook silently disables it;
- installation order changes behavior;
- every consumer must implement the same chaining protocol correctly;
- consumers need additional bookkeeping to associate hook state with individual windows.

Even our first-party `@remote-dom/core` integration assigns hooks directly, so another integration assigning the same operation can replace it.

The extension API moves composition into the polyfill. Extensions no longer need to know about each other or cooperate through manual callback chaining. It also separates two related responsibilities cleanly: the function body extends the available DOM APIs at construction time, while its return value subscribes to runtime DOM operations.

Returning a specialized constructor also keeps `new Window()` DOM-shaped and zero-argument:

```ts
const ExtendedWindow = Window.with(extension1, extension2);
const window = new ExtendedWindow();
```

### What this unlocks

This provides a foundation for optional DOM capabilities without growing the base polyfill for every use case:

- a complete `MutationObserver` implementation;
- CSSOM APIs such as `CSSStyleDeclaration` and `Element#style`;
- framework compatibility extensions like the CSSOM shims currently installed by `@remote-dom/react`;
- custom renderers, synchronization layers, instrumentation, and other hook consumers that can coexist safely;
- independently published polyfill extensions;
- eventually expressing the `@remote-dom/core` synchronization hooks as a regular extension.

Each extended `Window` remains reusable, while every instance receives independent installation state.

### What’s deliberately not included?

This PR establishes the extension and composition model. It does not yet:

- migrate `@remote-dom/core` to an extension;
- implement a complete `MutationObserver`;
- add late installation through `window.use()`;
- define uninstall or window-disposal semantics;
- expose creation hooks for the initial `html`, `head`, and `body`.

Extensions run after the base window and its initial document are constructed. This gives each extension a fully initialized window and matches the existing behavior of `window[HOOKS]`, which also does not observe document bootstrap.

### Changes since initial review

- Simplified extensions from `{name, install()}` objects to plain functions.
- Removed name-based duplicate detection; window APIs are last-write-wins, while hooks are additive.
- Switched the hook dispatcher from arrow closures to bound functions.
- Documented construction-time setup, runtime subscriptions, composition rules, and `window[HOOKS]` as legacy.
- Rebased onto `main`, preserving its `EVENT_PHASE_*` constants alongside extension hook dispatch.

### Testing

- Added coverage for reusable and chained extended constructors.
- Covered every existing DOM hook operation.
- Covered deterministic ordering and legacy `window[HOOKS]` compatibility.
- Covered independent per-window state, API-only extensions, and later extensions overriding APIs installed by earlier ones.
- Full test suite, type-checking, linting, bundle-size checks, Playwright E2E tests, and Web Platform Tests pass in CI.
- Included a minor changeset for `@remote-dom/polyfill`.

### DIFF
--- a/packages/polyfill/README.md
+++ b/packages/polyfill/README.md
@@ -45,17 +45,14 @@ import {Window, type WindowExtension} from '@remote-dom/polyfill';
 
 class CustomMutationObserver {}
 
-const mutationObserverExtension: WindowExtension = {
-  name: 'mutation-observer',
-  install(window) {
-    window.MutationObserver = CustomMutationObserver;
+const mutationObserverExtension: WindowExtension = (window) => {
+  window.MutationObserver = CustomMutationObserver;
 
-    return {
-      setAttribute(element, name, value) {
-        // Notify observers associated with this window.
-      },
-    };
-  },
+  return {
+    setAttribute(element, name, value) {
+      // Notify observers associated with this window.
+    },
+  };
 };
 
 const ExtendedWindow = Window.with(mutationObserverExtension);
@@ -68,9 +65,9 @@ are called in the same order for each DOM operation.
 Extensions are installed after the base window and its initial document have
 been constructed, so their hooks do not observe the document’s bootstrap.
 
-For low-level integrations, hooks can also be assigned directly through the
-exported `HOOKS` symbol. Extension hooks run first, followed by the hook
-assigned through `window[HOOKS]`:
+Assigning hooks directly through the exported `HOOKS` symbol is the legacy
+integration API. New integrations should use extensions instead. Extension
+hooks run first, followed by the legacy hook assigned through `window[HOOKS]`:
 
 ```ts
 import {HOOKS} from '@remote-dom/polyfill';
--- a/packages/polyfill/README.md
+++ b/packages/polyfill/README.md
@@ -37,8 +37,15 @@ This process will install polyfilled versions of the following globals:
 ## Extensions
 
 Use `Window.with()` to create a reusable `Window` subclass with additional DOM
-APIs or behavior. Each extension installs itself on every new window and can
-return hooks that subscribe to DOM operations:
+APIs or behavior. An extension is a function with two distinct parts:
+
+- **The function body runs once per window, during construction.** It receives
+  the new `Window` instance and can install or replace APIs on it, like
+  `window.MutationObserver`. Use it for setup, not for reacting to DOM changes.
+- **The returned hooks run for the lifetime of the window.** They subscribe to
+  DOM operations on that window, such as creating elements, setting attributes,
+  and adding event listeners, and are called each time one of those operations
+  happens. Returning hooks is optional.
 
 ```ts
 import {Window, type WindowExtension} from '@remote-dom/polyfill';
@@ -46,8 +53,10 @@ import {Window, type WindowExtension} from '@remote-dom/polyfill';
 class CustomMutationObserver {}
 
 const mutationObserverExtension: WindowExtension = (window) => {
+  // Construction time: extend this window instance.
   window.MutationObserver = CustomMutationObserver;
 
+  // Runtime: subscribe to DOM operations on this window.
   return {
     setAttribute(element, name, value) {
       // Notify observers associated with this window.
@@ -59,8 +68,14 @@ const ExtendedWindow = Window.with(mutationObserverExtension);
 const window = new ExtendedWindow();
 ```
 
-Extensions are installed in the order passed to `Window.with()`. Their hooks
-are called in the same order for each DOM operation.
+Extensions are installed in the order passed to `Window.with()` and across
+chained `.with()` calls. The two parts compose differently:
+
+- **Window APIs override.** Assignments like `window.MutationObserver = …` are
+  plain property writes, so when multiple extensions set the same API, the last
+  extension installed wins.
+- **Hooks are additive.** Every installed extension’s hooks are called in
+  installation order for each DOM operation.
 
 Extensions are installed after the base window and its initial document have
 been constructed, so their hooks do not observe the document’s bootstrap.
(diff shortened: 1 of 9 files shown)


packages/polyfill/source/tests/extensions.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+import {HOOKS} from '../constants.ts';
+import type {WindowExtension} from '../extensions.ts';
+import type {Hooks} from '../hooks.ts';
+import {Window} from '../Window.ts';
+
+describe('Window extensions', () => {
+  it('creates reusable Window subclasses', () => {
+    const installs: string[] = [];
+    const first = extension('first', () => {
```
vividviolet: nit: the `extension` helper doesn't do much and make the test harder to read. Can we just inline it?
```ts
const first = {name: 'first', install: () =>  { installs.push('first'); }};
```

packages/polyfill/README.md
```diff
@@ -34,12 +34,48 @@ This process will install polyfilled versions of the following globals:
...
+  install(window) {
+    window.MutationObserver = CustomMutationObserver;
+
+    return {
+      setAttribute(element, name, value) {
+        // Notify observers associated with this window.
+      },
+    };
+  },
+};
```
andrewiggins: `{name: string, install()}` kind of looks like an object pretending to be a function (`{name: string, apply(...parms)}`). Reminds me of webpack plugins a little bit.

What if we just made this a function and Window invoked those to create the plugin hooks?

```ts
export type WindowExtension = (
  window: Window,
) => Partial<Hooks> | void;

function mutationObserverExtension(window: Window) {
  let observers = new Set<MutationObserver>();

  window.MutationObserver = createMutationObserver(observers);

  return {
    setAttribute(element, name, value) {
      // Notify this window’s observers.
    },
  };
}

const ExtendedWindow = Window.with(mutationObserverExtension);
```

vividviolet (approved):
Looks good!  Just a minor nit

packages/polyfill/source/Window.ts
```diff
@@ -168,3 +190,47 @@ export class Window extends EventTarget {
     Object.defineProperties(globalThis, eventTargetPrototypeProperties);
   }
 }
+
+function installExtension(window: Window, extension: WindowExtension) {
+  if (window[EXTENSIONS].some(({name}) => name === extension.name)) {
```
andrewiggins: Could we rely on on identity instead of name here? Do names need to be globally unique?

packages/polyfill/README.md
```diff
@@ -34,12 +34,48 @@ This process will install polyfilled versions of the following globals:
...
+
+Extensions are installed in the order passed to `Window.with()`. Their hooks
+are called in the same order for each DOM operation.
+
+Extensions are installed after the base window and its initial document have
+been constructed, so their hooks do not observe the document’s bootstrap.
+
+For low-level integrations, hooks can also be assigned directly through the
+exported `HOOKS` symbol. Extension hooks run first, followed by the hook
+assigned through `window[HOOKS]`:
```
andrewiggins: Can we say this is the legacy way of registering hooks so future engineers and agents don't use this any more? I'd prefer extensions be the way forward to installing hooks.

packages/polyfill/source/Window.ts
```diff
@@ -168,3 +190,47 @@ export class Window extends EventTarget {
...
+    removeAttribute: (...args) => dispatchHook(window, 'removeAttribute', args),
+    createText: (...args) => dispatchHook(window, 'createText', args),
+    setText: (...args) => dispatchHook(window, 'setText', args),
+    insertChild: (...args) => dispatchHook(window, 'insertChild', args),
+    removeChild: (...args) => dispatchHook(window, 'removeChild', args),
+    addEventListener: (...args) =>
+      dispatchHook(window, 'addEventListener', args),
+    removeEventListener: (...args) =>
+      dispatchHook(window, 'removeEventListener', args),
+  };
```
andrewiggins, line 204: I have a vague memory of us doing some benchmarking on `.bind` vs `() => {}` in signals/preact and I think we settled on `.bind` since it doesn't capture scope. I wonder if we should use `.bind` here for partial application instead of arrow functions. For example:

```ts
{
   createElement: dispatchHook.bind(null, window, 'createElement');
}
```

That being said, without benchmarks (working on it), it may not matter that much

packages/polyfill/README.md
olavoasantos: making it an object allow us to add other things if we need to later (eg dependencies, teardown, etc). But we're not there yet and, if we ever do get there, we can accept both objects and functions and avoid breaking changes. I'll refactor.

packages/polyfill/source/Window.ts
olavoasantos: yea. initially this was meant to avoid having duplicates (eg two mutation observer extensions). But honestly this doesn't add much aside from noise. I'll clean up.

vividviolet: Hmm actually it does seem like there's 2 behaviours:
- globals override each other
- hooks are additive

I think that's fine but we should make it clear in the README

olavoasantos: good point. i'll adjust the readme.

vividviolet (approved):

andrewiggins (approved):

PR #662, opened 2026-09-02
Fix event listener registration semantics

olavoasantos:
## Problem

`EventTarget` tracked listener registrations in a way that could conflate event types and capture modes, and it did not retain enough registration-specific state to remove `once` listeners and `AbortSignal`-bound listeners reliably. Duplicate registrations could run more than once; an already-aborted signal could still register a listener; and removal through the wrong capture mode could orphan a once wrapper.

## Impact

**Important.** Event handlers are part of the polyfill’s public DOM surface. Incorrect listener identity can duplicate application work or leave callbacks active after callers explicitly remove or abort them, which makes worker-side event behavior diverge from the browser contract.

## Reproduction

```ts
const target = new EventTarget();
const listener = vi.fn();

target.addEventListener('change', listener, {once: true});
target.removeEventListener('change', listener, true);
target.removeEventListener('change', listener);
target.dispatchEvent(new Event('change'));
```

Before this change, the wrong-capture removal could make the once wrapper unremovable, so the final dispatch invoked `listener`. It now invokes no listener. The regression tests also show that the same callback is deduplicated by type, callback, and capture; that once registrations remain distinct across event types; and that already-aborted signals do not register listeners while a later abort removes an active one.

## Change

Store registrations per event type and per capture/bubble phase, keyed by the original callback. Each registration retains its normalized callback and, where applicable, its abort handler. Removal now operates on that exact registration, allowing `once` and signal cleanup to use the capture value recorded at registration time. Listener dispatch reads the corresponding structured type/phase list, avoiding the previous event-name key collision.

## Tests

`packages/polyfill/source/tests/event-target.test.ts` adds regression coverage for duplicate registration, separate capture registrations, colliding event names, once-listener identity and removal, already-aborted and active `AbortSignal` lifecycles, mutated options objects, and stale abort handlers after a listener is re-registered.

## Stack

- Stack: **Events** (#695), layer 1 of 3
- Base: `main`

## Validation

Fresh GitHub CI on restacked head `e249e3e` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;

### DIFF
--- a/.changeset/fair-listeners-register.md
+++ b/.changeset/fair-listeners-register.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/polyfill': patch
+---
+
+Correct event listener identity and AbortSignal handling in the EventTarget polyfill.
--- a/packages/polyfill/source/Event.ts
+++ b/packages/polyfill/source/Event.ts
@@ -13,8 +13,6 @@ export const EVENT_PHASE_BUBBLING = 3;
 
 export type EventPhase = number;
 
-export const CAPTURE_MARKER = '@';
-
 export interface EventInit {
   bubbles?: boolean;
   cancelable?: boolean;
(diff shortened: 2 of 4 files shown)


henrytao-me (approved):
LGTM

PR #664, opened 2026-09-02
Correct event dispatch lifecycle and listener snapshotting

olavoasantos:
## Problem

A dispatched `Event` retained transient propagation state after completion, the same instance could be redispatched while it was already in flight, and iteration over the live listener set allowed additions made during dispatch to run immediately. `composedPath()` also exposed the mutable internal path array.

## Impact

**Important.** Reused events could carry stale propagation state into a later dispatch, nested dispatch could violate the event lifecycle, and a listener could change which callbacks execute in the same delivery. Those outcomes make observable event ordering and state unreliable in the worker polyfill.

## Reproduction

```ts
const target = new EventTarget();
const addedListener = vi.fn();

target.addEventListener('change', () => {
  target.addEventListener('change', addedListener);
});

target.dispatchEvent(new Event('change'));
```

Before this change, iteration over the live listener collection could invoke `addedListener` during that first dispatch. It now waits until the next dispatch. The companion lifecycle tests verify that in-flight redispatch throws `InvalidStateError`, a completed event can be dispatched again with propagation state reset, and completed events expose `NONE`, `null`, and an empty composed path for phase, current target, and path respectively.

## Change

Add an internal dispatching guard and wrap dispatch in `try`/`finally` cleanup. Each dispatch resets propagation flags before building its path and normalizes transient state afterward. Listener delivery snapshots registrations, skips ones removed or aborted after the snapshot, and `composedPath()` returns a defensive copy so listener code cannot alter traversal state.

## Tests

`packages/polyfill/source/tests/event-dispatch.test.ts` covers propagation-state reset across redispatch, in-flight redispatch with and without a global `DOMException`, additions and remove/re-adds during dispatch, listeners removed or aborted during a snapshot, defensive composed paths, and post-dispatch state normalization.

## Stack

- Stack: **Events** (#695), layer 3 of 3
- Base: `fix-event-cancellation`
- Review after PR #663.

## Validation

Fresh GitHub CI on restacked head `66c608f` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me:
The lifecycle cleanup and registration snapshots look good. One correction before merging: preserve propagation flags explicitly set before dispatch.

At reviewed head `9829505e`, [the entry resets](https://github.com/Shopify/remote-dom/blob/9829505e18f8d4ddf56e766d1a4a2657c016c917/packages/polyfill/source/EventTarget.ts#L173-L175) erase an intentional stop:

```ts
const target = document.createElement('div');
let calls = 0;
target.addEventListener('change', () => calls++);

const event = new Event('change');
event.stopPropagation();
target.dispatchEvent(event);

console.log(calls);
```

- Native DOM-node dispatch: `0`.
- Actual base (`44e3b367`, #663): `0`.
- This PR: `1`.

The same introduced first-delivery regression occurs with `stopImmediatePropagation()` and `cancelBubble = true` for non-capture listeners. Pre-stopped capture listeners were already incorrect at the base, so that sibling case is not an introduced regression.

The [DOM invocation algorithm](https://dom.spec.whatwg.org/#concept-event-listener-invoke) returns when the stop-propagation flag is set. [Dispatch clears propagation flags at completion](https://dom.spec.whatwg.org/#concept-event-dispatch), not at entry.

Suggested correction:
- Remove the entry resets for `cancelBubble` and `STOP_IMMEDIATE_PROPAGATION`; retain the `finally` cleanup.
- Check the stop-propagation flag before invoking a listener group, not between listeners in the same group. Removing the resets alone would still leave the pre-existing capture case incorrect.
- Add focused tests for all three stop forms with capture/non-capture listeners. Assert the first dispatch invokes no listeners, cleanup resets transient state, and a second dispatch invokes the listener once. Keep the existing stopped-during-dispatch reuse tests green.

I ran all 32 event tests successfully against a pinned-source snapshot; the existing tests do not cover this case. The reproduction was checked against native Chrome DOM elements and the actual base/head. This was not a live Shell tophat.

olavoasantos:
yea, this is valid. i removed the entry resets for `cancelBubble` and `STOP_IMMEDIATE_PROPAGATION`, and `fireEvent()` now checks propagation state before entering each listener group. The new coverage exercises `stopPropagation()`, `stopImmediatePropagation()`, and `cancelBubble = true` for capture and non-capture listeners, including cleanup and reuse on a second dispatch. i force-updated the head, so this needs a fresh look.

henrytao-me (approved):

PR #665, opened 2026-09-02
Give MutationObserver its standard method surface

olavoasantos:
## Summary

- Give MutationObserver its standard method surface.
- Add focused regression coverage for the corrected behavior.

## Stack

This draft is part of the Globals stack. Its base is `main`.

## Validation

This reconstructed subsystem stack passes build, lint, forced type-check, and the full unit test suite.


henrytao-me (changes requested):
There is already Polyfill here https://github.com/Shopify/remote-dom/pull/624 

We should close this PR right?

PR #667, opened 2026-09-02
Validate tree mutations before changing node links

olavoasantos:
## Problem

`ParentNode` insertion and replacement operations could change links before confirming that the requested mutation was valid. Inserting a node into itself or one of its descendants could create a cycle, while an invalid reference node could detach an otherwise valid child before throwing. Identity cases also had incorrect mutation semantics: `insertBefore(node, node)` could throw and `replaceChild(nextSibling, oldChild)` could detach the sibling being installed.

## Impact

**Critical.** A cyclic parent/child graph can make later tree walkers hang or overflow. A rejected insertion or replacement could also corrupt the local tree before reporting the error, leaving callers with a changed DOM after an operation that failed.

## Reproduction

```ts
const child = document.createElement('span');
parent.appendChild(child);

parent.appendChild(parent);
```

Before this change, the final call could create a cycle. It now throws while `parent` remains detached and still contains `child`. The regression suite also demonstrates that `parent.insertBefore(child, child)` is a no-op and that replacing an old child with its next sibling leaves that sibling installed exactly once.

## Change

Validate the reference-node and hierarchy constraints before detaching or relinking anything. The hierarchy walk recognizes template-content host relationships as well as ordinary parents, so inserting a template into its own content is also rejected. After validation, identity insertion is a no-op and replacement uses the validated successor reference before unlinking the old child.

## Tests

`packages/polyfill/source/tests/parent-node-mutation-validation.test.ts` adds regression coverage for:

- self and ancestor insertion without state changes;
- invalid foreign references while preserving the attached source child;
- same-node insertion and replacement no-ops;
- replacement with the next sibling without detaching it; and
- DocumentFragment and template-content cycles, including mutually nested templates.

## Stack

- Stack: **Core foundation** (#697), layer 1 of 10
- Base: `main`

## Validation

Fresh GitHub CI on restacked head `f4cd474` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


henrytao-me (approved):
LGTM

JoviDeCroock (approved):

henrytao-me:
The validation-before-mutation changes look good, but could we keep the observable mutation effects instead of treating identity operations as complete no-ops?

At `f4cd474a`, the early returns in [replaceChild()](https://github.com/Shopify/remote-dom/blob/f4cd474aa4d870a04cf7328755996a1c9047ac95/packages/polyfill/source/ParentNode.ts#L106-L115) and [insertIntoValidated()](https://github.com/Shopify/remote-dom/blob/f4cd474aa4d870a04cf7328755996a1c9047ac95/packages/polyfill/source/ParentNode.ts#L147-L148) preserve the final tree but suppress native lifecycle callbacks and MutationObserver records.

Concrete repro in a fresh document/registry:

```js
const calls = [];
class ReviewItem extends HTMLElement {
  connectedCallback() { calls.push('connected'); }
  disconnectedCallback() { calls.push('disconnected'); }
}
customElements.define('review-identity-item', ReviewItem);
const child = document.createElement('review-identity-item');
parent.appendChild(child);
document.body.appendChild(parent);
calls.length = 0;

const observer = new MutationObserver(() => {});
observer.observe(parent, {childList: true});
parent.replaceChild(child, child);

console.log(calls);
console.log(observer.takeRecords().map(record => ({
  added: record.addedNodes.length,
  removed: record.removedNodes.length,
})));
observer.disconnect();
```

- Native Chrome and the actual base both report `['disconnected', 'connected']` and `[{added: 0, removed: 1}, {added: 1, removed: 0}]`.
- This head reports `[]` and `[]`. The final tree is unchanged in all three.

That makes self-replacement an introduced regression, rather than a pre-existing gap. Native `insertBefore(child, child)` also emits those callbacks and records; the base already detached/threw for that operation, while this head fixes its final tree but drops its effects.

Suggested narrow correction:
- Remove the identity early returns.
- Normalize a reference equal to the moving node to that node's next sibling, then continue through the mutation path. Keep validation before changing links.
- Add connected custom-element and exact MutationObserver assertions for self-insertion/self-replacement, retaining the existing invalid-operation and final-tree checks.

The DOM [pre-insert](https://dom.spec.whatwg.org/#concept-node-pre-insert) and [replace](https://dom.spec.whatwg.org/#concept-node-replace) algorithms normalize the reference rather than declaring the operation an observable no-op. This does not require pulling broader reaction timing, return-value, or error-name work into this PR.

I missed this in my earlier tree-only pass. All 109 polyfill tests pass in the pinned-source snapshot, including the 15 new cases, but the identity tests currently only assert final-tree state. Five focused structural comparisons still match native on this head. No live Shell tophat performed.

olavoasantos:
yea, those early returns preserved the final shape but dropped the mutation effects. `insertBefore(node, node)` now normalizes the reference to `node.nextSibling`, and `replaceChild(node, node)` continues through removal and insertion. The tests assert the disconnect/connect callbacks and exact remove/add `MutationRecord`s while keeping the final position unchanged.

PR #668, opened 2026-09-02
Preserve NamedNodeMap attribute invariants

olavoasantos:
## Problem

`NamedNodeMap.setNamedItem()` could corrupt its linked attribute list and ownership metadata. Reinstalling an attribute already present at the head or in the middle could truncate the attributes that followed it. Replaced and removed attributes could retain stale `ownerElement` or next-link state, and the same `Attr` instance could be installed on two elements at once.

## Impact

**Important.** Attribute enumeration and lookups could silently lose attributes after a valid-looking `setNamedItem()` call. Stale ownership and cross-element aliasing also mean a later mutation of one `Attr` can affect or misrepresent another element’s attribute state.

## Reproduction

```ts
const first = new Attr('first', 'one');
const second = new Attr('second', 'two');
const third = new Attr('third', 'three');

element.attributes.setNamedItem(first);
element.attributes.setNamedItem(second);
element.attributes.setNamedItem(third);
element.attributes.setNamedItem(second);
```

Before this change, reinstalling the middle `second` attribute could discard `third` from the list. It now returns the existing attribute and preserves the ordered list as `first`, `second`, `third`. Installing an attached `Attr` on a different element now throws `InUseAttributeError` until the attribute is removed from its original owner.

## Change

Treat installation of the identical attribute instance as an idempotent operation rather than rewiring the list. Reject attributes already owned by another element, preserve the replacement position when swapping an attribute, and clear both owner and next-link metadata when attributes are removed or replaced.

## Tests

`packages/polyfill/source/tests/named-node-map.test.ts` covers reinstalling an existing attribute at the head, middle, and tail; replacement and removal cleanup at every list position; and cross-element reuse before and after removal.

## Stack

- Stack: **Core foundation** (#697), layer 2 of 10
- Base: `validate-tree-mutations`
- Review after PR #667.

## Validation

Fresh GitHub CI on restacked head `afa837d` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me:
Could we bring forward the ownership/link cleanup ordering already present in #669, without bringing the whole reaction queue into this PR?

At `afa837d1`, [removal](https://github.com/Shopify/remote-dom/blob/afa837d186ee1fc41f97ba28b034f72810a84f2e/packages/polyfill/source/NamedNodeMap.ts#L81-L88) clears `NEXT`/`OWNER_ELEMENT` after the callback and hook, and [replacement](https://github.com/Shopify/remote-dom/blob/afa837d186ee1fc41f97ba28b034f72810a84f2e/packages/polyfill/source/NamedNodeMap.ts#L149-L164) similarly clears the replaced attribute's owner afterward. Reentrant callbacks can restore or transfer the attribute before that cleanup runs.

Concrete restoration repro in a fresh registry:

```js
let savedAttr;
class RestoringAttribute extends HTMLElement {
  static observedAttributes = ['state'];
  attributeChangedCallback(name, oldValue, newValue) {
    if (newValue === null) {
      this.attributes.setNamedItem(savedAttr);
    }
  }
}
customElements.define('review-restoring-attribute', RestoringAttribute);
const element = document.createElement('review-restoring-attribute');
element.setAttribute('state', 'old');
savedAttr = element.attributes.getNamedItem('state');
element.removeAttribute('state');

console.log(element.attributes.getNamedItem('state') === savedAttr);
console.log(savedAttr.ownerElement === element);
```

Native Chrome and the actual base report `true, true`. This head reports `true, false`: the callback reinstalls the attribute, then the outer cleanup clears its newly valid ownership. The same issue occurs when restoring a replaced attribute.

Related cases from the same cleanup order:
- Transferring the removed/replaced attribute to another element inside the callback succeeds natively and on the base, but the new guard throws `InUseAttributeError` here because it sees the stale source owner.
- A throwing callback skips cleanup, so subsequent reuse is incorrectly rejected by that guard.

The callback-visible stale ownership was pre-existing, but the final-owner corruption and new reuse failures are introduced by these changes.

Suggested narrow fix:
- Finalize the detached attribute's `NEXT`/`OWNER_ELEMENT` while committing list changes, before callbacks or hooks. Preserve the old value needed for notifications.
- Bring forward the cleanup placement from [#669's NamedNodeMap](https://github.com/Shopify/remote-dom/blob/19051d7b449083961aecab2256ef934d8e5e98e9/packages/polyfill/source/NamedNodeMap.ts), leaving its broader reaction queue there.
- Add focused removal/replacement tests for callback restoration, transfer to another element, and throwing callbacks, asserting exact final list/ownership and valid reuse.

The DOM [attribute removal algorithm](https://dom.spec.whatwg.org/#concept-element-attributes-remove) also clears ownership before handling attribute changes. I verified that #669 fixes these ownership cases; changing exception propagation is not part of this request.

The ordinary list fixes look good: all 117 polyfill tests passed in the pinned snapshot, and 20 structural native comparisons plus same-Attr callback/observer behavior matched. The six callback scenarios above expose the missing cleanup ordering. No live Shell tophat performed.

olavoasantos:
This ordering is already carried by #669: removal clears `NEXT` and `OWNER_ELEMENT` before hooks/reactions, and replacement clears the old attribute owner before notification. Since both layers are landing in the same release, i'd keep this PR focused and avoid duplicating the same transition here.

PR #669, opened 2026-09-02
Make custom-element reactions reentrant-safe

olavoasantos:
## Problem

Custom-element lifecycle and attribute callbacks ran inline with DOM mutations. A `connectedCallback()` that removed its own element could therefore emit the removal hook before the original insertion hook, even though the element finished detached. Callback errors could be thrown after local links changed but before their Remote DOM hook was emitted. Reentrant attribute callbacks had the same ordering problem and could leave the hook stream reporting a stale final value. Connectivity was also applied one node at a time, allowing callbacks to observe a partially connected subtree.

## Impact

**Critical.** Remote DOM hooks are the synchronization boundary between the worker-side polyfill and its receiver. Reentrant callbacks could make that hook stream describe the opposite order from the final local tree, and callback failures could suppress an already-earned synchronization effect. This can leave lifecycle behavior and rendered state unreliable.

## Reproduction

```ts
class SelfRemovingElement extends HTMLElement {
  connectedCallback() {
    this.remove();
  }
}

customElements.define('self-removing', SelfRemovingElement);
const element = document.createElement('self-removing');
document.body.appendChild(element);
```

The regression test in `packages/polyfill/source/tests/custom-element-reactions.test.ts` records the hooks around this operation. Before this change, it observed `remove` before `insert` while `element` finished detached. It now observes `insert`, then `remove`, matching the committed mutation order and final tree. The same suite verifies that an attribute callback changing `state` from `outer` to `inner` emits `outer`, then `inner`.

## Change

Introduce a FIFO custom-element reaction queue around parent and attribute mutations. Each mutation commits links and full-subtree connectivity, records any MutationObserver effects, and emits the corresponding Remote DOM hook before lifecycle or attribute reactions run. Nested mutations append their reactions to the same queue, so they cannot overtake effects earned by the outer mutation; queued callbacks continue to drain before the first callback error is rethrown.

## Tests

`packages/polyfill/source/tests/custom-element-reactions.test.ts` exercises:

- self-removal and self-reinsertion from lifecycle callbacks, with hook ordering and final connectivity;
- complete subtree connectivity before connected or disconnected callbacks run;
- throwing lifecycle and attribute callbacks, including preservation of committed local state and hooks;
- replacements, moves, and DocumentFragment insertion under reentrancy; and
- reentrant attribute updates and attribute restoration after removal.

## Stack

- Stack: **Core foundation** (#697), layer 3 of 10
- Base: `fix-named-node-map-invariants`
- Review after PR #668.

The restack preserves upstream MutationObserver records alongside deferred custom-element reactions.

## Validation

Fresh GitHub CI on restacked head `19051d7` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me:
The hook-order and full-subtree connectivity fixes look good. Could we adjust the reaction queue so nested public DOM mutations run their affected reactions before returning to the calling callback?

At `19051d7b`, [isPerformingMutations remains true during the flush](https://github.com/Shopify/remote-dom/blob/19051d7b449083961aecab2256ef934d8e5e98e9/packages/polyfill/source/custom-element-reactions.ts#L10-L32). A mutation invoked from a callback takes the `return mutation()` shortcut and leaves its reactions pending until that callback returns.

Concrete repro in a fresh registry:

```js
class ReadyElement extends HTMLElement {
  static observedAttributes = ['state'];
  ready = false;

  connectedCallback() {
    this.setAttribute('state', 'ready');
    console.log(this.ready);
  }

  attributeChangedCallback() {
    this.ready = true;
  }
}
customElements.define('review-ready-element', ReadyElement);
document.body.appendChild(document.createElement('review-ready-element'));
```

- Native Chrome and the actual base print `true`.
- This head prints `false`; the attribute callback runs only after `connectedCallback()` returns. The attribute value itself is already updated.

This is an introduced synchronous timing regression, not an async race: the delayed reaction still runs within the outer insertion call. The same regression appears with `appendChild()` inside `connectedCallback()` and `setAttribute()` inside another attribute callback.

Suggested direction:
- Separate internal mutation batching from callback execution. Keep committing the local state and Remote DOM hooks before invoking reactions.
- Use scoped element queues and per-element reaction ordering, so a public mutation invoked by a callback flushes its affected reactions before returning to that callback.
- Do not just toggle the global flag or give each call an isolated callback array; nested flushing must preserve pending outer work and existing reactions for the same element.

A useful regression case for that last point: A and B already have connection reactions queued, and A's callback sets an observed attribute on B. Native logs:

```text
A:start
B:connected
B:attribute
A:after-set
```

The head logs `A:after-set` before both B reactions. B's pending connection must run before its new attribute reaction, and both must run before the nested setter returns.

Please add immediate-after-call assertions for nested set/append/remove and attribute-in-attribute, plus that pending-sibling case. The existing ancestor-removes-child test checks callback order but has no marker after `remove()`, so it misses the return boundary. Preserve the existing exact hook-order, connectivity, ownership, and error-recovery tests.

The HTML [CEReactions steps](https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions) push a queue, run the operation, pop/invoke reactions, then return. The [reaction stack](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-reactions-stack) also maintains per-element reaction queues rather than a single global callback FIFO.

All 131 polyfill tests and 20 additional queue/error-recovery assertions passed in the pinned-source checks; native comparisons exposed five nested-timing differences. Callback exceptions escaping mutation methods instead of being reported like native errors are pre-existing and are not the new timing finding. No live Shell tophat performed.

olavoasantos:
Agreed. The global FIFO couldn't preserve the return boundary for nested public mutations or the pending reaction order for the same element. i replaced it with scoped per-element reaction queues and moved composite `ParentNode` operations onto internal mutation primitives, so nested `setAttribute()`, append, and removal reactions flush before returning to the calling callback without replaying the current reaction.

PR #672, opened 2026-09-02
Preserve Document structure in textContent assignments

olavoasantos:
## Problem

`Document` inherited generic parent-node `textContent` behavior. Its getter returned an empty string, and assigning any value used the generic child-replacement path, which could remove the initialized `documentElement` even though `document.head` and `document.body` still referenced the removed nodes.

## Impact

**Minor.** Code that reads `document.textContent` received the wrong value, while an assignment could leave the polyfill's initialized document shell structurally inconsistent. Subsequent code could retain apparently valid `head` and `body` references that were no longer connected to the document.

## Reproduction

```ts
const {documentElement, head, body} = document;

document.textContent = 'replacement';
```

Before this change, the generic setter could replace `documentElement` with a text node and detach `head` and `body`. It now leaves the initialized `documentElement`, `head`, and `body` in place; `document.textContent` returns `null`.

## Change

Override `Document.textContent` with document-specific behavior: return `null` from the getter and ignore assignments. This protects the initialized document structure without changing `textContent` mutation behavior for other parent nodes.

## Tests

Regression coverage in `packages/polyfill/source/tests/text-content.test.ts` verifies the getter and assignments of a string, empty string, and `null`. It asserts that the original document nodes remain linked and connected and that no create, insert, or remove hooks run. A companion case confirms that `body.textContent = 'replacement'` still performs the normal parent-node mutation.

## Stack

- Stack: **Core foundation** (#697), layer 6 of 10
- Base: `empty-parent-text-content`
- Review after PR #671.

## Validation

Fresh GitHub CI on restacked head `675216c` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me:
The document-structure protection and null getter look right. One small correction before this lands: the empty setter skips input conversion that native DOM still performs.

At `675216c39c3b9ee294ed460288319dd0fb44fded`, `Document.ts:59` is `set textContent(_data: any) {}`. Web IDL converts the assigned value to `DOMString?` before running the Document-specific no-op setter steps:

```js
const error = new Error('conversion failed');
let calls = 0;

document.textContent = {
  toString() {
    calls++;
    throw error;
  },
};
```

- Native Chrome and the actual #671 base: one conversion call, throw the original error, document tree preserved, zero mutation records.
- This head: zero conversion calls, no error, document tree preserved.

So this introduces a regression on the throwing-conversion path that #671 already handles correctly. Ordinary object conversion, `Symbol.toPrimitive` with the string hint, and conversion side effects are also skipped.

Suggested fix: keep the null getter and do not call the generic mutation setter, but perform nullable string conversion and discard the result. For example:

```ts
set textContent(data: any) {
  if (data !== null && data !== undefined) {
    void `${data}`;
  }
}
```

Template-literal conversion requests the string hint and rejects Symbols. `String(Symbol())` alone would accept them. Primitive Symbol acceptance was already wrong on the base; the newly skipped object conversion/errors are the regression here.

Please retain the existing structure/hook cases and add focused tests for:
- A converter called exactly once, including the `Symbol.toPrimitive` string hint.
- Original conversion errors propagating without changing the document or emitting setter-induced hooks/records.
- Symbol rejection while preserving document structure.

All 147 polyfill tests pass on this head; four of the five new cases fail on the actual base. Native comparisons exposed eight conversion differences across 20 scenarios. A probe-only coercion override matched all 20 native results; it was not an implementation edit or live Shell tophat.

References:
- https://webidl.spec.whatwg.org/#dfn-attribute-setter
- https://webidl.spec.whatwg.org/#js-DOMString
- https://dom.spec.whatwg.org/#dom-node-textcontent

olavoasantos:
Agreed. The no-op setter now still performs nullable string conversion with the string hint and discards the result. That preserves the document while keeping conversion side effects, original error identity, and `Symbol` rejection.

PR #673, opened 2026-09-02
Make wide-tree traversal stack-safe and insertion transactional

olavoasantos:
## Problem

Tree traversal used recursive calls for both descendants and sibling lists. A tree with roughly 9,000–10,000 siblings could therefore overflow the JavaScript call stack while collecting `textContent`, running selector queries, or updating subtree connectivity.

Insertion made this more dangerous: it could rewrite parent and sibling links before traversing the inserted subtree. If that traversal then failed, the local tree was partially changed but MutationObserver records, Remote DOM hooks, connectivity, and custom-element reactions were not completed consistently.

## Impact

**Important.** Wide or deeply nested content could throw a `RangeError` during ordinary DOM operations. During insertion or replacement, that failure could also leave the worker’s local DOM and the remote receiver describing different trees, making later rendering and lifecycle behavior unreliable.

## Reproduction

```ts
const root = document.createElement('div');

for (let index = 0; index < 10_000; index++) {
  root.appendChild(document.createTextNode('x'));
}

root.textContent;
```

Before this change, traversing a sufficiently wide sibling list could overflow the call stack. Selector queries over an equivalent tree had the same failure mode. Inserting a connected wide subtree could fail after structural links had already changed but before synchronization effects completed.

## Change

- Replace recursive descendant and selector walks with iterative, stack-safe traversal.
- Preflight every subtree traversal required for connectivity and lifecycle processing before detaching or attaching nodes.
- Commit links and connectivity only after preparation succeeds.
- Queue MutationObserver records, Remote DOM hook effects, and custom-element reactions from the committed tree in deterministic order.
- Drain nested hook effects FIFO so reentrant mutations cannot overtake effects already earned by the outer mutation.

The traversal and transactional changes belong together: making the walk stack-safe prevents the common overflow, while preflighting the walk guarantees that any other traversal failure leaves both local and remote state untouched.

## Tests

Regression coverage includes:

- 10,000-wide text and selector traversal;
- a chain deeper than 10,000 nodes;
- connectivity, hooks, and lifecycle reactions for wide subtree insertion;
- fault-injected insertion, replacement, and fragment traversal failures that must leave links and hooks unchanged;
- transactional MutationObserver records for fragments, moves, replacements, and rejected insertions; and
- reentrant and throwing hook scenarios that verify effect and reaction ordering.

## Stack

- Stack: **Core foundation** (#697), layer 7 of 10
- Base: `document-text-content-semantics`
- Review after PR #672.

The restack integrates upstream MutationObserver records into the validate → prepare → commit pipeline and adds focused fragment, move, replacement, and rejected-preflight coverage.

## Validation

Fresh GitHub CI on restacked head `2275003` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me:
The iterative traversal and preflight changes work, but the deferred hook queue introduces a remote-tree sync regression. FIFO ordering of live-node callbacks is not sufficient when an insertion hook serializes the subtree at dispatch time.

Reproduced on `2275003933dca3be9848c01830250d0e818193e1` against actual base `675216c39c3b9ee294ed460288319dd0fb44fded`, using the real `@remote-dom/core` polyfill hooks, `RemoteReceiver`, and a synchronous receiver subscription. No mocked serializer or receiver.

The sequence is:

```js
fragment.append(first, second);

// root is already connected to receiver.connection.
let once = true;
receiver.subscribe(receiver.root, () => {
  if (once && receiver.root.children.length === 1) {
    once = false;
    second.appendChild(added);
  }
});

root.appendChild(fragment);
```

When the first insertion is delivered, the subscriber edits the second subtree before its outer insertion has been published:

- Base: the nested insert hook sees second as remotely disconnected and sends nothing. The later second insertion serializes added once. Local and receiver both have `second = [added]`.
- Head: the nested hook is queued. The outer second insertion serializes its now-current subtree, including added, then the queued nested insertion sends added again. Local has `[added]`; receiver has `[added, added]`, including a duplicate remote ID.

There is also a data-loss variant: start second with `[removed, kept]` and remove removed in the subscription. Head serializes second as `[kept]`, then replays the queued removal at index 0 and deletes kept. Local remains `[kept]`; receiver becomes `[]`. Base stays synchronized.

The mismatch is between `ParentNode.ts:403-430`, which queues closures over live nodes, and `packages/core/source/polyfill/hooks.ts:22-36`, which calls `serializeRemoteNode(node)` only when that delayed hook runs. The pending insertion can already include a nested edit that is then replayed as a separate effect.

Suggested correction: make queued effects and insertion snapshots represent the same mutation boundary. Prepare immutable transport payloads before reentrant dispatch, or coalesce edits already represented by pending subtree publication. Please keep the stack-safe traversal/preflight improvements; merely deduplicating receiver IDs would not fix the removal case. The preparation/dispatch contract needs to account for the hook consumers, not only callback order.

Please add real core hook/receiver integration tests for pending-subtree append and removal, exact local/receiver tree equality, and subsequent edits. Controls that mutate an already-published subtree or add a root sibling stay synchronized on both versions; the existing FIFO-order tests miss the pending-subtree serialization overlap.

Validation: all 169 polyfill tests and 22 additional hook-queue assertions pass. The core integration probe stays synchronized in 4/4 cases on base, 2/4 on head. This is synchronous in-process reentrancy, not an asynchronous race or live Shell tophat. No implementation fix has been validated yet.

Sources:
- https://github.com/Shopify/remote-dom/blob/2275003933dca3be9848c01830250d0e818193e1/packages/polyfill/source/ParentNode.ts#L403-L430
- https://github.com/Shopify/remote-dom/blob/2275003933dca3be9848c01830250d0e818193e1/packages/core/source/polyfill/hooks.ts#L22-L36

olavoasantos:
yea, queuing closures over live nodes was the problem. Hook effects now track pending subtree publications and suppress nested hooks already represented by that pending snapshot. Once a publication starts, later edits emit normally. The coverage includes the real core hooks and `RemoteReceiver` for both duplicate-append and removal/data-loss cases, plus subsequent edits.

PR #674, opened 2026-09-02
Update attached attribute values atomically

olavoasantos:
## Problem

An `Attr` installed on an element did not receive that element's owner document. Updating its `value` or `nodeValue` first changed the stored value, then attempted to read hooks through the missing document context and threw.

## Impact

**Important.** Direct mutation of an attached attribute could report a failure after local state had already changed and before the Remote DOM attribute hook was dispatched. That leaves the local element and its remote receiver able to disagree about the attribute value.

## Reproduction

```ts
const attribute = new Attr('data-state', 'initial');

element.attributes.setNamedItem(attribute);
attribute.value = 'updated';
```

Before this change, the final assignment changed the attribute locally but threw while resolving its hook context. It now completes, exposes `'updated'` through both `attribute.value` and `element.getAttribute('data-state')`, and dispatches the attribute hook.

## Change

Set an attribute's owner document when `NamedNodeMap` attaches it to an element. In the attached-attribute setter, retain the existing conditional MutationObserver record path, update the value before hook dispatch, and use the attached attribute's document hooks. Detached or removed attributes continue to update only their own value; reattachment changes the hook destination to the new document.

## Tests

`packages/polyfill/source/tests/attr-mutation.test.ts` covers both `value` and `nodeValue` on an attached attribute, including state visibility when the hook runs. It also verifies detached and removed attributes do not dispatch hooks, and that a removed attribute reattached to an element in another `Window` dispatches only through that window's hooks.

## Stack

- Stack: **Core foundation** (#697), layer 8 of 10
- Base: `stack-safe-wide-traversal`
- Review after PR #673.

The restack preserves upstream attribute MutationObserver records and the value commit → observer record → Remote DOM hook ordering.

## Validation

Fresh GitHub CI on restacked head `2d276b9` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me (approved):
LGTM for the scoped owner-document/hook-context fix. Attached Attr.value/nodeValue updates now resolve the correct document hooks, preserve value -> observer record -> hook ordering, and synchronize ordinary HTML attributes with the core receiver. Removed attributes remain local-only, and reattachment switches the hook destination correctly.

Non-blocking follow-up: direct Attr.value/nodeValue assignments still skip attributeChangedCallback. This is pre-existing on the actual base, not a regression introduced here, so it does not need to be folded into this PR.

It matters for RemoteElement because the core polyfill hooks intentionally skip custom-element attributes and rely on attributeChangedCallback for synchronization. With remoteAttributes = ['data-state'], I reproduced:

```js
// element is a connected RemoteElement with data-state="initial".
const attribute = element.attributes.getNamedItem('data-state');
attribute.value = 'updated'; // Same gap with nodeValue.

// Local: "updated"; receiver: "initial".
// Control: element.setAttribute('data-state', 'final') syncs both to "final".
```

For that follow-up, please share the attribute-change/reaction handling with NamedNodeMap, preserve the installed Attr identity, and capture the old value even when no MutationObserver is active. Simply routing through element.setAttribute() would replace the Attr object in this implementation. Add value/nodeValue tests for callback old/new values and exact core receiver state without requiring an observer.

Validation on 2d276b9: 174 polyfill tests, 56 focused core tests, and 27 additional hook assertions passed, plus native comparisons and actual core hook/receiver probes. This is scoped code-review approval, not live Shell tophat or full locked-build sign-off. The full local core suite was blocked by missing htm/mini and jsdom in the scratch toolchain.

Upstream corrections, including #673's deferred-hook serialization fix, still need to land before merging the integrated stack; please restack and rerun checks afterward.

Relevant paths:
- https://github.com/Shopify/remote-dom/blob/2d276b9f06adaf05cd01ce3c85a87df1b46ce3d4/packages/polyfill/source/Attr.ts#L47-L71
- https://github.com/Shopify/remote-dom/blob/2d276b9f06adaf05cd01ce3c85a87df1b46ce3d4/packages/core/source/polyfill/hooks.ts#L53-L64

PR #675, opened 2026-09-02
Remove slot attributes without resurrection

olavoasantos:
## Problem

Removing a `slot` attribute called the slot-property setter with an empty value as part of the attribute-change reaction. The setter reflected that value back into an attribute, so `removeAttribute('slot')` removed the attribute for the hook stream but recreated `slot=""` in the local tree.

## Impact

**Important.** The worker-side element could still contain `slot=""` after the remote receiver had processed an attribute removal. That local/host divergence affects slot assignment and makes later attribute or property updates unreliable.

## Reproduction

```ts

element.setAttribute('slot', 'aside');
element.removeAttribute('slot');
```

Before this change, the removal hook was emitted but the local element regained `slot=""`. It now has no `slot` attribute, `getAttribute('slot')` returns `null`, and `element.slot` remains the empty string. Assigning `element.slot` still reflects property-driven updates through the attribute.

## Change

Remove the special slot-property write from the attribute-reaction helper. Attribute removal now remains a removal; the slot property derives its empty value from an absent attribute instead of recreating one. This also lets a custom element callback deliberately restore the attribute only after the removal hook has been emitted.

## Tests

Polyfill regression coverage in `packages/polyfill/source/tests/slot-attribute.test.ts` covers empty and non-empty values, repeated removal, property-driven updates, and a custom-element callback that restores the attribute. Core coverage in `packages/core/source/tests/elements.test.ts` verifies that removal is reflected both locally and on the remote receiver, including the emitted attribute-property mutation.

## Stack

- Stack: **Core foundation** (#697), layer 9 of 10
- Base: `mutate-attached-attributes`
- Review after PR #674.

## Validation

Fresh GitHub CI on restacked head `917ee19` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me (approved):
LGTM for the scoped slot-removal fix. The slot getter already derives its value from the attribute, so removing the redundant property assignment prevents slot="" resurrection without breaking property-driven updates or deliberate callback restoration.

Validated on 917ee196: 179 polyfill tests and 59 focused core tests passed; all eight added tests fail on the actual base. All 20 native comparisons matched, and all six actual core hook/receiver scenarios produced the expected removal/restoration behavior.

This is scoped code-review approval, not live Shell tophat or full locked-build sign-off. Upstream corrections, including #673, still need to land; please restack and rerun checks before merging the integrated stack.

PR #676, opened 2026-09-02
Reject invalid and duplicate custom element definitions

olavoasantos:
## Problem

`CustomElementRegistry.define()` accepted invalid custom-element names and silently replaced an existing registration. It also allowed one constructor to be registered under more than one name.

## Impact

**Minor.** Invalid or duplicate definitions were accepted as though they succeeded, so application code could lose the original element definition or observe a registry state the browser would reject. Failing at registration time gives callers a deterministic error and preserves the previously valid definition.

## Reproduction

```ts
const registry = new CustomElementRegistryImplementation();
class OriginalElement extends HTMLElement {}
class ReplacementElement extends HTMLElement {}

registry.define('original-element', OriginalElement);
registry.define('original-element', ReplacementElement);
```

Before this change, the second definition was accepted and replaced the first one. It now throws a `NotSupportedError`, and `registry.get('original-element')` remains `OriginalElement`. Invalid names such as `element`, `My-element`, and reserved names such as `font-face` now throw `SyntaxError` instead of being registered.

## Change

Validate names against the polyfill’s current registration grammar and the reserved-name set before mutating the registry. Reject both an already-defined name and a constructor already present in the registry with `NotSupportedError`. Validation occurs before storage or `whenDefined()` notification, preserving the established registration and pending listeners after rejection.

This is scoped to the package’s exposed validation contract; it does not claim complete browser potential-custom-element-name conformance.

## Tests

`packages/polyfill/source/tests/custom-element-registry.test.ts` covers accepted ASCII and Unicode examples under this implementation; rejected uppercase, whitespace, control-character, separator, and reserved names; duplicate names; duplicate constructors; original-registration preservation; and ensuring `whenDefined()` listeners run only after a successful definition.

## Stack

- Stack: **Globals and custom-element registry** (#707), layer 2 of 2
- Base: `install-dom-globals-consistently`
- Review after PR #666.

## Validation

Fresh GitHub CI on restacked head `16fb020` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


JoviDeCroock (approved):

henrytao-me:
The `define()` checks look good. I'd make one small consistency correction: reuse the name validator in `whenDefined()` too.

At reviewed head `16fb0201`, [`whenDefined()`](https://github.com/Shopify/remote-dom/blob/16fb02016c7301e6a8e22966edf58e780b41b447/packages/polyfill/source/CustomElementRegistry.ts#L78-L92) still allocates pending waiters for names that `define()` now rejects:

```ts
registry.whenDefined('element');
registry.whenDefined('My-element');
registry.whenDefined('font-face');
```

Native Chrome returns a Promise rejected with `SyntaxError` for each. This PR leaves each Promise pending, and no successful `define()` can fulfill it.

To be precise, the missing rejection in `whenDefined()` is pre-existing, not a new regression in that method. On the actual base (`072b7e8a`, #666), an incorrectly accepted `define('element', Constructor)` could fulfill the waiter. With this PR, that definition correctly throws, leaving the waiter impossible to fulfill through the public API.

Suggested correction:
- Extract one name-validity predicate, including the reserved-name check, and share it between `define()` and `whenDefined()`.
- For invalid names, `whenDefined()` should return a Promise rejected with a `SyntaxError` DOMException, not throw synchronously. Validate before allocating a waiter.
- Add focused tests for ordinary invalid names, uppercase names, and reserved names. Preserve valid pending/already-defined behavior and the existing successful-definition notification test.

This matches the [HTML `whenDefined()` algorithm](https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-whendefined) without expanding into a broader registry rewrite.

Validation: all 135 polyfill tests passed against the pinned-source snapshot. Native comparisons matched the tested `define()` name and duplicate checks; the three discrepancies were these invalid-name waiters. No live Shell tophat performed.

olavoasantos:
Agreed. `define()` and `whenDefined()` now share the same name predicate. Invalid names return a Promise rejected with a `SyntaxError` `DOMException` before we allocate a waiter; valid pending and already-defined behavior stays unchanged. i force-updated the head, so this needs a fresh look.

PR #677, opened 2026-09-02
Normalize DOM names and attribute namespaces

olavoasantos:
## Problem

The polyfill applied name and namespace rules inconsistently across its exposed DOM APIs. Generic `Node.nodeName` uppercased special node names; HTML custom-element lookup used the raw `createElement()` argument rather than its normalized name; HTML attribute names were stored and hooked without ASCII folding; and namespaced attributes were matched by qualified name rather than namespace plus local name.

## Impact

**Important.** Consumers could fail to instantiate a registered custom element, miss an HTML attribute through a differently cased lookup, receive inconsistent names in Remote DOM hooks and custom-element callbacks, or fail to read, replace, and remove a namespaced attribute through its expanded name. SVG names and casing could also be represented incorrectly.

## Reproduction

```ts
class ExampleElement extends HTMLElement {}
customElements.define('x-example', ExampleElement);

const element = document.createElement('X-EXAMPLE');
element.setAttribute('DATA-STATE', 'ready');
element.setAttributeNS('urn:state', 'first:state', 'initial');

const state = element.getAttributeNS('urn:state', 'state');
```

Before this change, raw custom-element lookup could miss `x-example`, HTML attribute casing was not consistently normalized, and a namespace-aware lookup by local name could not find `first:state`. It now creates the registered element, stores and exposes `data-state` consistently, and returns `'initial'` for the expanded-name lookup. Non-element nodes retain names such as `#text`, while SVG element names preserve their case.

## Change

- Normalize and validate element and attribute names at the relevant DOM API boundaries, including qualified-name namespace rules.
- Track element and attribute prefixes, derive local names, and match namespaced attributes by namespace plus local name while preserving the stored qualified name.
- ASCII-fold HTML element and non-namespace-aware attribute operations, but preserve SVG and namespace-aware casing.
- Route attribute hooks and custom-element reactions through the normalized local name and namespace. Existing MutationObserver record queuing remains in the attribute mutation paths; this commit does not add observer-specific tests.
- Apply the same normalization and hook/reaction path to `toggleAttribute()`, which this commit adds as a related API extension.

`toggleAttribute()` is scope beyond the original audit finding; it is included here so this newly exposed name-based attribute operation follows the corrected normalization rules.

## Tests

Focused coverage adds or extends tests for special-node names; HTML, SVG, and qualified element names; custom-element lookup; attribute storage, lookup, hooks, and reactions; expanded-name replacement and removal; valid and invalid namespace/name combinations; selector and `getElementsByTagName()` ASCII behavior; and `toggleAttribute()` semantics. The `getElementsByTagName()` cases also promote the corresponding WPT capabilities from deferred to supported.

## Stack

- Stack: **Core foundation** (#697), layer 10 of 10
- Base: `remove-slot-without-resurrection`
- Review after PR #675.

The restack preserves upstream attribute MutationObserver records alongside namespace-aware custom-element reactions.

## Validation

Fresh GitHub CI on restacked head `2a58ded` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


packages/polyfill/source/NamedNodeMap.ts
```diff
@@ -135,7 +167,7 @@ export class NamedNodeMap {
     let prev: Attr | null = null;
 
     while (child) {
-      if (child.name === attr.name && child[NS] == attr[NS]) {
+      if (child.localName === attr.localName && child[NS] === attr[NS]) {
```
JoviDeCroock, line 170: I reproduced replacing new Attr('first:state', 'same', 'urn:state') with new Attr('second:state', 'same', 'urn:state') via setNamedItemNS(): the local name changes, but no hook fires. With different values, only the new name is set—the old name is never removed. Since the remote hook adapter keys attributes by qualified name, the receiver can retain stale attributes.

henrytao-me:
Reviewed `2a58dedaeab47c1fdc1bbb05f39d013fcfc0d6b2` against its actual #675 base, `917ee19632491a994eebf9a33512f8e48b394c67`. The normalization and expanded-name fixes make sense, but I reproduced three issues to correct before approval.

### 1. Keep null-namespace semantics in the consumers of qualified-name lookup

`NamedNodeMap.getNamedItem()` now correctly searches qualified names across namespaces. However, `Element.id`/`slot`, ID/class/attribute selector matchers, and `getElementsByClassName()` still use `getAttribute()`/`hasAttribute()` assuming null namespace.

```js
element.setAttributeNS('urn:foreign', 'id', 'ghost');
document.body.appendChild(element);

console.log(element.id);
console.log(document.getElementById('ghost') === element);
console.log(document.querySelector('#ghost') === element);
```

Native and the base produce `''`, `false`, `false`. This head produces `'ghost'`, `true`, `true`. Similar probes expose foreign `class`/`slot` values and false `[data-flag]` matches. When foreign attributes are installed first, they can hide later ordinary attributes.

Writes also need the same audit: setting `element.id` after a foreign `id` now modifies that foreign attribute instead of creating the null-namespace attribute. Setting `slot` to the foreign value can skip ordinary attribute creation; setting another value creates the ordinary attribute but the getter still reads the foreign one.

Please retain the corrected qualified-name lookup and make these semantic consumers explicitly use null-namespace access. Preserve HTML ASCII normalization for selector attribute names. Tests should cover foreign-only attributes, both installation orders, reflected writes, and exact lookup/selector results.

### 2. Convert the value before validating names/namespaces

```js
const original = new TypeError('conversion failed');
const value = {toString() { throw original; }};

element.setAttribute('bad name', value);
// Also: element.setAttributeNS(null, 'prefix:state', value);
```

Native/base call `toString()` once and throw `original`. This head skips conversion and throws `InvalidCharacterError` / `NamespaceError` instead. Side effects from a successful value conversion are skipped too.

Please stage argument conversions once, in Web IDL order, before name/namespace validation and mutation. Keep the new validation: the base incorrectly accepted invalid names when value conversion succeeded. Add original-error identity, conversion count/order, and side-effect tests for both setters.

Source: https://github.com/Shopify/remote-dom/blob/2a58dedaeab47c1fdc1bbb05f39d013fcfc0d6b2/packages/polyfill/source/Element.ts#L120-L149

### 3. Qualified-name replacement must update both transport keys

I also confirmed [Jovi's existing finding](https://github.com/Shopify/remote-dom/pull/677#discussion_r4016749613) with the actual core polyfill hooks and `RemoteReceiver`, not mocked serialization.

Replacing an installed `first:state` Attr with a new `second:state` Attr in the same namespace through `setNamedItemNS()` produces:

- Equal values: local has only `second:state`; receiver has only `first:state` because no hook runs.
- Different values: local has only `second:state`; receiver retains both qualified names.
- Subsequent update and removal still leave stale `first:state` remotely.

The base mirrored its incorrect two-attribute local state; this PR correctly replaces the local Attr but introduces receiver divergence. Please remove the old transport key and publish the replacement key when the qualified name changes, even if the value is unchanged. Keep the correct expanded-name replacement semantics and add exact real-receiver assertions through replacement, update, and removal.

### Positive follow-up and validation

This PR already includes the shared Attr reaction handling mentioned in my #674 approval. Direct `Attr.value` / `nodeValue` now synchronize custom `RemoteElement` attributes while preserving the installed Attr identity and capturing old values without an observer. The six HTML/custom setter probes improve from 4/6 synchronized on #675 to 6/6 here, so no duplicate follow-up implementation is needed for those cases.

- 226 polyfill tests and 59 focused core tests pass.
- Seven changed test files on the actual base: 46 fail / 46 pass, including inherited tests.
- 68 isolated native/base/head scenarios, plus actual core hook/receiver probes, separate these issues from inherited gaps and browser/spec differences.
- Locked Prettier passes; reported head CI is green.

This is scoped source/native/core validation using a borrowed test runner, not a full locked build or live Shell/admin tophat. Upstream corrections and restacking remain separate gates.

olavoasantos:
Fixed all three:

- `id`, `slot`, selectors, class lookup, `setAttribute()`, `toggleAttribute()`, and `removeAttribute()` now use explicit null-namespace access where their contracts require it, while generic qualified-name lookup remains unchanged.
- `setAttribute()` and `setAttributeNS()` stage value conversion before name/namespace validation.
- expanded-name replacement now removes the old qualified transport key and publishes the new key even when the value is unchanged.

The transport case is covered through the real core hooks and `RemoteReceiver`, including later update and removal.

packages/polyfill/source/NamedNodeMap.ts
olavoasantos: yea, confirmed. When the expanded name matches but the qualified name changes, we now remove the old transport key and set the new one even when the value is unchanged. The local list still performs one expanded-name replacement, and the receiver tests cover equal/different values, later update, and removal.

JoviDeCroock (approved):

henrytao-me (approved):

PR #678, opened 2026-09-02
Respect node namespace context in isDefaultNamespace

olavoasantos:
## Problem

`Node.isDefaultNamespace()` previously returned whether its argument was the HTML namespace, without consulting the receiver. Consequently, it reported the same result for an HTML element, an SVG element, a descendant, a detached attribute, and a node with an inherited or overridden `xmlns` declaration.

## Impact

**Minor.** The exposed namespace API returned an incorrect boolean for non-HTML and null default-namespace contexts, so namespace-aware consumers could make decisions from the document-wide HTML constant rather than the node’s actual namespace context.

## Reproduction

```ts
const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
const text = document.createTextNode('text');
svg.appendChild(text);

svg.isDefaultNamespace('http://www.w3.org/2000/svg');
text.isDefaultNamespace('http://www.w3.org/2000/svg');
```

Before this change, both calls returned `false` because only the HTML namespace compared equal. They now return `true`. The regression tests also cover document and HTML contexts, prefixed elements with inherited and overridden `xmlns` declarations, empty declarations normalized to `null`, and attached versus detached attributes.

## Change

Resolve the receiver’s default namespace with DOM locate-a-namespace semantics: unprefixed elements use their namespace URI; prefixed elements consult a default `xmlns` declaration and ancestors; documents start at `documentElement`; attributes use their owner element; and document fragments, document types, and detached attributes have no default namespace. Normalize the requested namespace before comparing it to that resolved context.

## Tests

Adds `is-default-namespace.test.ts` coverage for HTML, SVG, descendant, document, attribute, inherited/default-override, and empty-`xmlns` cases, including matching and non-matching namespace arguments.

## Stack

- Stack: **Namespace and selectors** (#698), layer 1 of 3

## Validation

Fresh GitHub CI on restacked head `5100d3f` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


marvinhagemeister (approved):

JoviDeCroock (approved):

henrytao-me:
The receiver-aware namespace lookup is a useful fix. I reviewed head `5100d3f` against its actual #677 base `2a58ded`; two corrections are still needed:

### 1. Convert the argument before reading namespace context

[Node.ts:67-68](https://github.com/Shopify/remote-dom/blob/5100d3f68dc1d61c44aef58803e1477cbcc3e7aa/packages/polyfill/source/Node.ts#L67-L68) evaluates `locateNamespace(this)` before `normalizeNamespace(namespace)`. Argument conversion can move the receiver or change the namespace declaration:

```js
const HTML = 'http://www.w3.org/1999/xhtml';
const SVG = 'http://www.w3.org/2000/svg';
const html = document.createElement('div');
const svg = document.createElementNS(SVG, 'svg');
const text = document.createTextNode('x');
html.appendChild(text);

text.isDefaultNamespace({
  toString() {
    svg.appendChild(text);
    return HTML;
  },
});
```

Native returns `false`; this head returns `true`. Both convert once and finish with the text under SVG, but this head compares the old HTML context. I also reproduced the opposite move, an `xmlns` update, and an Attr detachment inside conversion.

Please normalize the argument once into a local variable first, then resolve the receiver's namespace. [Web IDL argument conversion](https://webidl.spec.whatwg.org/#es-operations) precedes the [isDefaultNamespace algorithm](https://dom.spec.whatwg.org/#dom-node-isdefaultnamespace). The old constant-only implementation did not coerce objects at all, so this is not a claim that its conversion path was conformant.

### 2. Walk ancestors iteratively

[Node.ts:181-212](https://github.com/Shopify/remote-dom/blob/5100d3f68dc1d61c44aef58803e1477cbcc3e7aa/packages/polyfill/source/Node.ts#L181-L212) recurses once per unresolved ancestor:

```js
const HTML = 'http://www.w3.org/1999/xhtml';
let current = document.createElement('div');
for (let i = 0; i < 10000; i++) {
  const next = document.createElementNS('urn:own', 'p:item');
  current.appendChild(next);
  current = next;
}
current.isDefaultNamespace(HTML);
```

Native and the actual base return `true`; this head throws `RangeError`. I reproduced this with native DOM and bundled base/head polyfills inside the same Chrome 152, not only across different Node/browser stack limits. A 6,000-depth control succeeds.

Please use an iterative lookup while preserving document/attribute dispatch and fragment/doctype stopping rules. Add exact regression checks for conversion-driven context changes and the deep-chain result; these are not covered by the five added tests.

Validation: 231 polyfill tests, 59 focused core tests, the polyfill type-check with locked TypeScript 7.0.2, and locked formatting pass. All five added tests fail the actual base. A probe-only staged-conversion/iterative-lookup replacement fixes the five targeted scenarios without changing the other comparison results; no production source was changed. This is not full-repository or live Shell/admin sign-off, and #677's separate corrections still apply.

olavoasantos:
Fixed both. `isDefaultNamespace()` now normalizes its argument before resolving receiver context, so conversion-driven moves and namespace changes are observed. `locateNamespace()` is iterative now and preserves the existing document, attribute, fragment, and doctype rules; the 10,000-deep regression case completes without a `RangeError`.

PR #681, opened 2026-09-02
Correct ChildNode replacement and insertion order

olavoasantos:
## Problem

Upstream PR #623 already landed the baseline `replaceWith()` repair: ordinary replacement no longer passes the old and new nodes backwards, and zero- and multi-argument calls have their basic placement semantics. The remaining issue was that `replaceWith()`, `before()`, and `after()` still changed the tree one argument at a time. A later cyclic node or throwing string conversion could leave earlier nodes moved and the receiver removed; custom-element reactions could also observe or reenter an incomplete multi-node operation.

## Impact

**Important.** A throwing convenience mutation could leave connected trees partially changed, with sibling links, lifecycle callbacks, and hook effects reflecting only a prefix of the requested operation. This is especially hazardous for a worker-side DOM that must stay coherent with its remote receiver.

Upstream #623 landed the baseline fix while this stack was being prepared, narrowing this PR to the remaining ordering, validation, atomicity, and reentrancy cases.

## Reproduction

```ts
const receiver = document.createElement('em');
const replacement = document.createElement('span');
const source = document.createElement('div');

parent.appendChild(receiver);
source.appendChild(replacement);
document.body.append(parent, source);

receiver.replaceWith(replacement, parent);
```

The second replacement is the receiver’s parent and is invalid. Before this change, the operation could remove `receiver` and move `replacement` before reaching the invalid argument. It now throws before mutating either tree: `parent` still contains `receiver` and `source` still contains `replacement`.

## Change

Validate every node argument before changing links, then convert the complete argument list into one node or a document fragment and commit it in one parent operation. Anchor around arguments that are already adjacent to the receiver so their supplied order is preserved. Wrap `replaceWith()`, `before()`, and `after()` in the custom-element reaction boundary, ensuring all structural changes are committed before reactions can run or reenter.

This PR deliberately builds on #623 rather than redescribing its already-landed baseline fix; its scope is the remaining argument validation, ordering, atomicity, and reentrancy behavior.

## Tests

New regression coverage verifies `replaceWith()` with mixed nodes and strings, adjacent/self receiver arguments, an invalid first or later cyclic argument, and a reentrant disconnect callback that observes the complete replacement before removing one replacement node. Companion `before()` and `after()` coverage verifies argument order, complete-tree visibility to connected callbacks, late conversion failures without moves or hooks, lifecycle-error behavior, and late cyclic arguments without partial mutation.

## Stack

- Stack: **Tree and adoption** (#699), layer 1 of 2

## Validation

Fresh GitHub CI on restacked head `6676e1d` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


packages/polyfill/source/ChildNode.ts
```diff
@@ -46,3 +65,35 @@ export function toNode(parent: ParentNode, node: Node | any) {
...
+    if (!(node instanceof Node)) continue;
+
+    let ancestor: Node | null = parent;
+    while (ancestor) {
+      if (ancestor === node) {
+        throw Error(
+          'cannot insert a node into itself or one of its descendants',
+        );
+      }
+      ancestor = ancestor.parentNode;
```
JoviDeCroock: Could this use the same host-inclusive ancestor walk as ParentNode.validateInsertion (PARENT ?? HOST). That would add automatic support for `template.content`

olavoasantos:
Agreed. `before()`, `after()`, and `replaceWith()` now stage every `Node` / string conversion once before hierarchy validation, while fragment assembly and tree mutation still happen afterward. The combined throwing-conversion + later-cycle cases preserve the original trees and throw the original conversion error.

packages/polyfill/source/ChildNode.ts
olavoasantos: yea, switched the ancestor walk to `PARENT ?? HOST` so the prevalidation catches cycles through `template.content` before any earlier argument is moved.

packages/polyfill/source/ChildNode.ts
```diff
@@ -12,37 +13,94 @@ export class ChildNode extends Node {
...
+function convertNodesIntoNode(
+  parent: ParentNode,
+  nodes: (Node | string)[],
+): Node {
+  const convertedNodes: Node[] = [];
+  for (const node of nodes) convertedNodes.push(toNode(parent, node));
+  if (convertedNodes.length === 1) return convertedNodes[0]!;
+
+  const fragment = parent.ownerDocument.createDocumentFragment();
+  for (const node of convertedNodes) fragment.appendChild(node);
```
andrewiggins: IIUC, won't this fire custom element reactions prematurely, when instead we would want to delay those until after this entire operation is done? I.e. if a custom element is already mounted, and then it is moved, it should never see this move to the fragment. My agent came up with this test case to demonstrate the issue:

 ```ts
it('commits replacements before disconnecting a moved custom element', () => {
  const receiver = document.createElement('em');
  const second = document.createElement('span');
  let observedChildren: unknown[] = [];

  class MovingElement extends PolyfillHTMLElement {
    disconnectedCallback() {
      observedChildren = [...parent.childNodes];
    }
  }

  polyfillWindow.customElements.define(
    'moving-element',
    MovingElement as unknown as CustomElementConstructor,
  );

  const first = document.createElement('moving-element');
  parent.appendChild(receiver);
  document.body.append(parent, first); // first must already be connected

  receiver.replaceWith(first, second);

  expect([...parent.childNodes]).toEqual([first, second]); // passes
  expect(observedChildren).toEqual([first, second]); // fails: [receiver]
});
 ```

Should we expose a private method on `ParentNode` that any public insertion method can use to do the actual insertion that takes an array of children to insert so we can avoid having to go through fragments all together? Just one thought off the top of my head.

olavoasantos: yea, this reproduced. `convertNodesIntoNode()` was staging through public `fragment.appendChild()`, so that nested reaction boundary flushed while the destination still contained `receiver`.

i kept the fragment conversion instead of adding the array-based mutation path, but routed both staging and the final commit through internal `ParentNode` primitives and collect hook effects until all links are committed. `before()`, `after()`, and `replaceWith()` now cover moved connected custom elements, and argument conversion happens before reading `parentNode` too. Your exact case passes on `9904cdd`, and fresh CI is green. i force-updated the head, so this needs another look.

henrytao-me (approved):

PR #682, opened 2026-09-02
Adopt full subtrees during cross-document insertion

olavoasantos:
## Problem

Cross-document insertion updated `ownerDocument` only on the inserted root. Descendants and attached attributes retained the source document, and initialized `HTMLTemplateElement.content` was outside the ordinary child tree, so later mutations could resolve hooks through the wrong `Window`.

## Impact

**Important.** After inserting a subtree into another document, a descendant attribute, text, or child mutation could be dispatched through the source window rather than the destination. The local tree would appear inserted while its Remote DOM effects were sent to the wrong receiver.

## Reproduction

```ts
const source = new Window().document;
const destination = new Window().document;
const root = source.createElement('section');
const child = source.createElement('span');

child.setAttribute('data-state', 'before');
root.appendChild(child);
destination.body.appendChild(root);

child.setAttribute('data-state', 'after');
```

Before this change, `root.ownerDocument` changed but `child` and its attached attribute still belonged to `source`, so the final mutation could use the source window’s hook set. It now adopts the root, descendants, and attached attributes into `destination` before later mutations run.

## Change

Build an iterative adoption snapshot before insertion commits, covering the subtree, attached attributes, and already initialized template-content subtrees. Apply that snapshot to the destination document during insertion, and share the same adoption routine with `Document.adoptNode()`. The snapshot deduplicates malformed template host/content cycles and preflights traversal so an error leaves links, ownership, connectivity, and hooks unchanged.

## Tests

Regression coverage uses two windows to verify nested elements, text, and attached attributes receive the destination owner document and route later attribute, text, and insertion effects only through destination hooks. It also covers initialized and untouched template content, `Document.adoptNode()`, document-fragment roots, malformed template cycles, and failed preflight preservation. The 6,000-deep initialized-template adoption case is iterative and has an explicit 20-second test budget.

## Stack

- Stack: **Tree and adoption** (#699), layer 2 of 2
- Base: `correct-child-replace-with`
- Review after PR #681.

The 6,000-deep adoption regression test has an explicit 20-second CI budget; the full coverage run passes.

## Validation

Fresh GitHub CI on restacked head `9788a1c` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


marvinhagemeister (approved):

henrytao-me:
The full-subtree insertion fix looks right at `9788a1c` against #681 `6676e1d`, but `Document.adoptNode()` needs an effects-ordering correction before landing.

[Document.ts:124-132](https://github.com/Shopify/remote-dom/blob/9788a1cefe7e087d9999403caf69a93a9fc9096f/packages/polyfill/source/Document.ts#L124-L132) collects the adoption snapshot, calls public `removeChild()`, then applies the snapshot. That removal runs hooks and custom-element reactions before returning, so user code can invalidate the snapshot's membership.

Minimal polyfill reproduction:

```js
const source = new Window();
const destination = new Window();

source.customElements.define(
  'x-moving',
  class extends source.HTMLElement {
    disconnectedCallback() {
      this.appendChild(source.document.createElement('span'));
    }
  },
);

const root = source.document.createElement('x-moving');
source.document.body.appendChild(root);
destination.document.adoptNode(root);

root.ownerDocument === destination.document; // true
root.lastChild.ownerDocument === destination.document; // false on head
```

The equivalent native case and the actual base finish with both nodes owned by destination. Head leaves the newly appended child source-owned underneath a destination-owned root because the child was absent from the snapshot.

The inverse also corrupts ownership: if `disconnectedCallback()` moves an existing child into a third document, native/base leave it owned by that third document, but head's saved snapshot overwrites its owner with destination even though its parent remains in the third document.

Both new failures also reproduce from the source Window's `removeChild` hook, without custom elements. Later attribute/text mutations actually route through the wrong Window's hook set. The old callback-visible source ownership was already incorrect on base; these final ownership corruptions are introduced by applying the frozen snapshot afterward.

Please retain the iterative snapshot and preflight, but commit detachment and adoption before either removal hooks or lifecycle callbacks run, matching the insertion path's committed-state-before-effects ordering. Deferring only custom-element reactions misses hook reentrancy; resnapshotting after callbacks retains the wrong callback-visible context. The DOM operation is [CEReactions] and completes its [adoption steps](https://dom.spec.whatwg.org/#concept-node-adopt) before the callback boundary.

Focused regression coverage:
- Append a source-created child during removal, through both `disconnectedCallback` and `removeChild` hooks.
- Transfer an existing child into a third document through both paths.
- Assert exact final parent/owner consistency and later attribute/text hook routing, not just root ownership.
- Preserve explicit `adoptNode()` preflight failures: original error, unchanged parent/owners/connectivity, and no hooks. The current head passes the template-content traversal-error control.

Validation: 247 polyfill tests and 59 focused core tests pass; eight of nine added tests fail the actual base. The locked-TypeScript polyfill check, formatting, and reported CI pass. Isolated Chrome/native comparisons match all 40 ordinary operation/connectivity combinations (49/58 overall). Eight targeted hook scenarios expose four new explicit-adoption routing failures while confirming two implicit-insertion routing fixes. Existing inert-template-document and other adoptNode contract gaps are separate, not requests to expand this correction. No production/candidate fix, complete locked-repository/full-build validation, real multi-receiver integration, live Shell/admin validation, or approval performed.

olavoasantos:
Agreed. `adoptNode()` now preflights the snapshot, detaches the root, and commits ownership before removal hooks or disconnected reactions run. Reentrant appends are adopted by the destination operation, and transfers into a third document are no longer overwritten by a stale snapshot. The tests cover both hook and callback paths plus later hook routing.

andrewiggins (approved):

henrytao-me (approved):

PR #685, opened 2026-09-02
Preserve character data and handle void elements in innerHTML parsing

olavoasantos:
## Problem

The limited `innerHTML` parser discarded literal ampersands, did not decode character references, and stored encoded attribute values such that serialization could escape them again. It also treated HTML void elements as normal parents, nesting following content beneath them and emitting invented closing tags.

## Impact

**Important.** Ordinary valid HTML assigned through the exposed `innerHTML` path could lose text data, produce a different tree, or serialize to different markup. In a worker-side Remote DOM tree, that means consumers can make decisions from content that does not match the markup they assigned.

## Reproduction

```ts

element.innerHTML = '<p>Fish & chips</p><img src="logo.png"><p>After</p>';
```

Before this change, the literal `&` was dropped and the second paragraph was parsed beneath `img`, which was serialized with a closing tag. It now retains `Fish & chips`, keeps `img` and both paragraphs as siblings, and serializes the void element as ``.

## Change

- Decode the parser's supported named and numeric character references while preserving literal ampersands and unrecognized references.
- Decode attribute references before storing them so serialization escapes the value once.
- Keep tokenizers local to each parse call, preventing nested `innerHTML` parsing from corrupting an outer parse.
- Treat the defined HTML void-element set as childless during parsing and serialization.

This is a targeted correction to the existing limited parser, not browser-grade entity coverage or malformed-markup recovery.

## Tests

Adds `packages/polyfill/source/tests/inner-html-parsing.test.ts`, covering literal ampersands; the supported named, decimal, and hexadecimal references; attribute round-tripping; reentrant parsing from an attribute callback; void-element sibling structure; self-closing void syntax; and closing-tag omission for every supported HTML void element.

## Stack

- Stack: **HTML serialization** (#701), layer 1 of 3

## Validation

Fresh GitHub CI on restacked head `1b7881e` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;


packages/polyfill/source/serialization.ts
```diff
@@ -15,47 +15,84 @@ import type {Comment} from './Comment.ts';
...
+    },
+  );
+}
+
+function isVoidElement(name: string) {
+  return VOID_ELEMENTS.has(name.toLowerCase());
+}
 
 export function parseHtml(html: string, contextNode: Node) {
+  const elementTokenizer =
```
marvinhagemeister: Nit: Hoisting regexes out of hot functions is typically good for perf.

marvinhagemeister (approved):

JoviDeCroock (approved):

olavoasantos:
Fixed both. Void classification now uses the element namespace and exact `localName`, so foreign and established mixed-case elements keep their children while prefixed HTML void elements still use their local name. Named references now match only the exact aliases we support; arbitrary casing stays literal while hexadecimal digits remain case-insensitive.

packages/polyfill/source/serialization.ts
olavoasantos: yea, hoisted the tokenizer regexes and switched the loops to `matchAll()` so the global regex state remains isolated during reentrant parsing.

henrytao-me (approved):

PR #689, opened 2026-09-02
Default CustomEvent detail to null when omitted

olavoasantos:
## Problem

`CustomEvent.detail` was assigned directly from the optional initialization dictionary. Constructing a `CustomEvent` without a `detail` property therefore exposed `undefined`, whereas the event’s default detail value is `null`.

## Impact

**Minor.** Consumers that distinguish an omitted payload from an explicitly supplied value can observe the wrong public event contract. In the worker polyfill, that makes `CustomEvent` behave differently from the DOM API it models.

## Reproduction

```ts
new CustomEvent('change').detail;
new CustomEvent('change', {}).detail;
```

Before this change, both expressions evaluated to `undefined`. They now evaluate to `null`. The regression test also shows that explicitly supplied `null`, `0`, `false`, `''`, and object values are preserved; an explicitly `undefined` `detail` is normalized to `null` by this implementation.

## Change

Normalize the constructor’s `detail` value to `null` when it is nullish. The same normalization is applied by the legacy `initCustomEvent()` path so both ways of creating a custom event expose the same default.

## Tests

Adds focused constructor coverage for an omitted initialization dictionary, a dictionary without `detail`, explicit `undefined`, and explicitly supplied falsy and object values.

## Stack

- Stack: **Independent quick wins** (#702), layer 2 of 4
- Base: `fix-node-contains`
- Review after PR #688.

## Validation

Fresh GitHub CI on restacked head `5f0b228` passes:

- lint;
- type-check and build;
- unit tests with coverage;
- bundle-size checks;

### DIFF
--- a/.changeset/grumpy-cycles-suffer.md
+++ b/.changeset/grumpy-cycles-suffer.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/polyfill': patch
+---
+
+Default omitted CustomEvent detail values to null.
--- a/packages/polyfill/source/CustomEvent.ts
+++ b/packages/polyfill/source/CustomEvent.ts
@@ -10,7 +10,7 @@ export class CustomEvent<T = any> extends Event {
 
   constructor(type: string, eventInitDict?: CustomEventInit<T>) {
     super(type, eventInitDict);
-    this.detail = eventInitDict?.detail as any;
+    this.detail = (eventInitDict?.detail ?? null) as any;
   }
 
   /**
@@ -25,6 +25,6 @@ export class CustomEvent<T = any> extends Event {
     detail?: T,
   ) {
     super.initEvent(type, bubbles, cancelable);
-    (this as any).detail = detail;
+    (this as any).detail = detail ?? null;
   }
 }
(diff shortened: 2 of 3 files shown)


marvinhagemeister (approved):

JoviDeCroock (approved):

henrytao-me (approved):
LGTM for the scoped default-detail fix, reviewed at `5f0b2281bddfc7ac37c64503f40f38f4f9b98243` against actual #688 base `3c57e5b6202f0779eb7ee99ad3298f6d17032c51`.

The constructor and legacy `initCustomEvent()` now use the DOM null default for omitted/undefined detail while preserving explicit falsy values and payload identity. I found no introduced blocker.

Validation:
- 108 polyfill tests and 56 focused core tests pass.
- The eight added constructor cases produce 3 failures / 5 passes on the base, matching the intended default changes.
- All 43 isolated native/base/head comparisons match the head, including legacy defaults, payload identity, detail getters, original getter errors, and listener delivery.
- Locked Prettier passes; reported current-head CI is green.

Non-blocking test suggestion: add `initCustomEvent()` cases for resetting a populated event with omitted/undefined detail and preserving falsy/object values. This method changed too, while the new repository tests exercise only the constructor; the separate native probe verified both paths.

This approval is scoped code-review sign-off, not a full locked build or live Shell/admin tophat. Restack/recheck as the parent lands; integrated stack and live-validation gates remain separate.

PR #692, opened 2026-09-02
Distinguish DOM mutation and selector errors

olavoasantos:
## Problem

The polyfill reported invalid tree mutations with generic `Error` objects and treated some malformed or unsupported selectors as non-matches. Callers could not reliably distinguish a missing child or reference node, a hierarchy violation, and invalid selector syntax from other failures or an ordinary empty result.

## Impact

**Minor.** Code using the exposed DOM-like API cannot make DOM-style error decisions when every mutation failure is a generic error or invalid selector input silently produces no match. Mutation failures also need to report the error before they move nodes or emit Remote DOM hooks, so the local tree and remote receiver remain synchronized.

## Reproduction

```ts
const foreignParent = document.createElement('section');
const foreignChild = document.createElement('span');
foreignParent.appendChild(foreignChild);

parent.removeChild(foreignChild);
parent.querySelectorAll('');
```

Before this change, the first operation threw a generic `Error`, and the empty selector returned an empty result. They now throw errors named `NotFoundError` and `SyntaxError`, respectively. The regression tests likewise verify `HierarchyRequestError` for an attempted ancestor insertion; in every mutation-error case, parent/child links and insert/remove hook calls remain unchanged.

## Change

- Add a shared DOM-exception factory that returns a named `DOMException` when available and a named `Error` fallback when it is not.
- Report `NotFoundError` for invalid child or reference nodes and `HierarchyRequestError` for invalid insertion hierarchy, with hierarchy validation taking precedence where both conditions apply.
- Stage and validate all arguments to variadic insertion APIs before converting values or moving a cross-document node, preserving state if validation or conversion fails.
- Reject empty, malformed, and unsupported selector syntax with `SyntaxError` rather than silently returning no matches; supported Unicode and double-hyphen identifiers and relative `:has()` selectors remain covered.
- Reuse the shared factory for existing named custom-element, namespace, document-cloning, and event redispatch errors so the fallback behavior is consistent.

## Tests

Adds focused error-contract coverage for invalid children and references, hierarchy violations (including template host relationships), all variadic insertion APIs, and conversion failures after a cross-document node has been staged. The tests assert error names, unchanged links, no hook calls, and no custom-element reactions. Selector coverage exercises malformed/unsupported syntax through the parser, `querySelector()`, and `querySelectorAll()`, and verifies the no-`DOMException` fallback.

## Stack

This draft is the final fan-in for the subsystem stacks. The live PR remains parked on `polyfill-correctness-prerequisites`; its historical head also contains a bundle-size gate update that was separately landed by PR #706. An unpublished maintainer-local reconstruction isolates this error-contract layer on the fully integrated post-#706 base, but it is supporting evidence rather than a reviewable branch. After the prerequisite stacks merge, rebase this PR onto `main` and rerun the complete validation suite.

## Validation

The live PR’s historical CI is not a current green validation signal: its lint, type-check, bundle-size, Playwright, and classified Web Platform Test checks passed, but its unit-test check failed.

An unpublished maintainer-local reconstruction passed:

- lint;
- type-check and build;
- unit tests with coverage (56 files / 665 tests);
- bundle-size checks;
- the classified Web Platform Test suite (4/4 files); and
- Playwright end-to-end tests (13/13).

These local results are supporting evidence only. Fresh GitHub CI is still required after the live PR is rebased.

### DIFF
--- a/.changeset/distinguish-dom-errors.md
+++ b/.changeset/distinguish-dom-errors.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/polyfill': patch
+---
+
+Throw named DOM errors for invalid tree mutations and selector syntax.
--- a/packages/polyfill/source/ChildNode.ts
+++ b/packages/polyfill/source/ChildNode.ts
@@ -1,5 +1,6 @@
-import {NEXT, PREV} from './constants.ts';
+import {HOST, NEXT, PARENT, PREV} from './constants.ts';
 import {performWithCustomElementReactions} from './custom-element-reactions.ts';
+import {createDOMException} from './dom-exception.ts';
 import type {ParentNode} from './ParentNode.ts';
 import {Node} from './Node.ts';
 
@@ -15,12 +16,13 @@ export class ChildNode extends Node {
     if (!parent) return;
 
     return performWithCustomElementReactions(() => {
-      validateNodesForInsertion(parent, nodes);
+      const staged = stageNodes(nodes);
+      validateInsertionNodes(parent, staged);
 
       let next = this[NEXT];
-      while (next && nodes.includes(next)) next = next[NEXT];
+      while (next && staged.includes(next)) next = next[NEXT];
 
-      const replacement = convertNodesIntoNode(parent, nodes);
+      const replacement = convertNodesIntoNode(parent, staged);
       if (this.parentNode === parent) {
         parent.replaceChild(replacement, this);
       } else {
@@ -34,12 +36,13 @@ export class ChildNode extends Node {
     if (!parent) return;
 
     return performWithCustomElementReactions(() => {
-      validateNodesForInsertion(parent, nodes);
+      const staged = stageNodes(nodes);
+      validateInsertionNodes(parent, staged);
 
       let previous = this[PREV];
-      while (previous && nodes.includes(previous)) previous = previous[PREV];
+      while (previous && staged.includes(previous)) previous = previous[PREV];
 
-      const node = convertNodesIntoNode(parent, nodes);
+      const node = convertNodesIntoNode(parent, staged);
       parent.insertBefore(node, previous ? previous[NEXT] : parent.firstChild);
     });
   }
@@ -49,12 +52,13 @@ export class ChildNode extends Node {
     if (!parent) return;
 
     return performWithCustomElementReactions(() => {
-      validateNodesForInsertion(parent, nodes);
+      const staged = stageNodes(nodes);
+      validateInsertionNodes(parent, staged);
 
       let next = this[NEXT];
-      while (next && nodes.includes(next)) next = next[NEXT];
+      while (next && staged.includes(next)) next = next[NEXT];
 
-      const node = convertNodesIntoNode(parent, nodes);
+      const node = convertNodesIntoNode(parent, staged);
       parent.insertBefore(node, next);
     });
   }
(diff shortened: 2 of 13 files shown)


olavoasantos:
i'm keeping this branch parked until the prerequisite stacks land, so i haven't updated this head. The structural nested-`:has()` rejection moved down to #679 where acceptance begins. EOF recovery, compound-selector grammar, the duplicate append/prepend validation, and the remaining named-error coverage will stay in the final #692 reconstruction on current `main`, followed by fresh full CI.

henrytao-me:
Thanks, keeping this parked until the prerequisites land and taking the structural nested-`:has()` guard from #679 makes sense. One addition to the [existing EOF correction](https://github.com/Shopify/remote-dom/pull/692#issuecomment-5670527381) for the final reconstruction: some of the new malformed-attribute fixtures also need recovery rather than `SyntaxError`.

```js
const root = document.createElement('section');
const child = document.createElement('div');
child.setAttribute('data-kind', 'item]');
root.appendChild(child);

root.querySelector('[data-kind="item]') === child;
// Native: true, despite the deliberately missing closing quote/bracket.
// Current head: SyntaxError. Base: no match.
```

I confirmed this in isolated Chrome 153 with `CSS1Compat`, against unchanged head `23f091e7` / base `7ec42505`. The [first two quoted-attribute fixtures](https://github.com/Shopify/remote-dom/blob/23f091e708427902e22692f52a296c415e82cd21/packages/polyfill/source/tests/dom-mutation-selector-errors.test.ts#L312-L331) and their nested `:has()` variants are accepted natively at EOF. Positive fixtures with the literal values `item]` and `item"]` actually select the element; these are not just silently ignored invalid selectors.

Please include string/attribute EOF recovery alongside the function recovery already discussed, and replace those error expectations with exact matching assertions through both query APIs plus parser coverage. Keep the genuinely malformed operator/trailing-junk cases rejected. Base's missing positive match is an inherited gap; head's new exception is separate. Deliberately unsupported CSS remains a different policy boundary.

Fresh validation of this unchanged head:
- 469 polyfill and 59 focused core tests pass locally, using borrowed Vitest 4.1.2; locked polyfill TypeScript and formatting checks pass. The green suite currently includes these incorrect EOF expectations.
- Seven public API error paths preserve the named fallback when `DOMException` is absent.
- Repeated 6,000-template construction remains about 1.8x base: 291/296/291ms versus 541/534/538ms, with serialization under 3ms. This confirms redundant traversal overhead, not reproduction of the historical five-second CI timeout.

The earlier compound-selector and duplicate single-node validation corrections still apply to the planned reconstruction. No need to duplicate the upstream nested-`:has()` implementation here. Fresh full CI on the final restacked head is still the gate; these local checks are not integrated-stack or live Shell validation.

olavoasantos:
yea, added string/attribute EOF recovery to the final #692 reconstruction checklist alongside function EOF recovery, compound-selector grammar, and duplicate single-node validation. The live branch stays untouched until the prerequisites land; then i'll rebuild it on current `main` and run fresh CI.

olavoasantos:
@henrytao-me, the final reconstruction is now on current `main` at `ee48d529`.

- supported `:has()` / `:not()` function EOF recovery is preserved;
- quoted string and attribute-block EOF recovery now matches the literal values in direct and nested selectors;
- invalid type/universal placement such as `[data-kind]div` and `:not(.missing)div` reports `SyntaxError`; and
- single-node `append()` / `prepend()` no longer add a duplicate wrapper preflight. The 6,000-template construction probe had a 245.4 ms median versus 252.5 ms on current `main`, while multi-argument validation remains intentionally atomic.

The reconstruction also routes `InUseAttributeError` through the shared named-error fallback. Fresh CI is green: 8/8 checks pass, including bundle size, Playwright, and classified WPT.

marvinhagemeister (approved):

andrewiggins (approved):

PR #693, opened 2026-09-02
Add indexed and named NamedNodeMap access

andrewiggins:
## What changed

Add live browser-style property access to the polyfilled `NamedNodeMap` without making it inherit from `Array`:

```ts
element.attributes[0];
element.attributes.id;
```

- Resolve indexed and named reads through one shared proxy in the prototype chain, preserving `NamedNodeMap` identity and avoiding a proxy allocation for every element.
- Match `getNamedItem()` by qualified name across namespaces while preserving methods and inherited properties.
- Parse only canonical ECMAScript array-index property names, rejecting fractions, aliases, and values outside the 32-bit array-index range before slower string coercion or linked-list traversal.
- Keep `Attr.ownerElement`, linked-list state, and `ownerDocument` correct when attributes are removed, replaced, directly adopted, or carried across documents with an adopted or inserted subtree. Directly adopted attributes are detached from their current element before their document changes.
- Reject attempts to attach an attribute already owned by another element.
- Add a minor `@remote-dom/polyfill` changeset.

## Design and performance

A constructor-returned proxy prototype was benchmarked first, but it substantially deoptimized existing methods and added about 32 bytes of retained V8 heap per materialized map. The shared prototype fallback restores ordinary `getAttribute`, mutation, `length`, `item`, iteration, and serialization paths to approximately the pre-proxy baseline while retaining indexed and named reads.

The strict index parser was benchmarked against simpler coercion and bitwise alternatives. `Number.isInteger()` plus explicit bounds was the clearest option and performed at least as well overall, while avoiding duplicate linked-list traversal for unusual property names.

The final size-limit run reports `@remote-dom/polyfill` at **7.05 kB Brotli** (**+218 B**, 3.19%) against an **8 kB** limit.

## Web Platform Tests

- Enable six supported cases in `dom/nodes/attributes-namednodemap.html`; defer two cases that still require `Document.createAttribute()`, with one also requiring an exposed `NamedNodeMap` constructor.
- Promote the `Document-getElementById.html` case that mutates an ID through `element.attributes[0].value`.

Current classified WPT result: **5/5 files passed** with no supported failures or inventory drift.

## Intentional scope

The shared prototype fallback provides ordinary indexed and named reads, but it is not a complete Web IDL legacy-platform-object implementation. This change does not add `has`, own-key/descriptor reflection, assignment/deletion restrictions, alternate `Reflect.get()` receiver behavior, or indexed precedence over properties placed directly on `NamedNodeMap.prototype`. In-use attributes currently throw a basic `Error` because the polyfill does not expose `DOMException`.

`Document.createAttribute()`, global `NamedNodeMap` exposure, complete namespace/local-name modeling, and HTML case-folding remain follow-up work.

## Validation

- `pnpm type-check`
- `pnpm exec vitest run --project polyfill --project wpt-runner` — 16 files, 136 tests passed
- `pnpm test:wpt` — 5/5 classified files passed
- `pnpm lint`
- `pnpm size`

### DIFF
--- a/.changeset/named-node-map-properties.md
+++ b/.changeset/named-node-map-properties.md
@@ -1,0 +1,5 @@
+---
+'@remote-dom/polyfill': minor
+---
+
+Add live indexed and named property access to `Element.attributes`, including correct attribute ownership when values are changed, removed, replaced, adopted, or moved between documents.
--- a/packages/polyfill/source/Document.ts
+++ b/packages/polyfill/source/Document.ts
@@ -1,6 +1,7 @@
 import {
   NS,
   NAME,
+  NODE_TYPE_ATTRIBUTE,
   NODE_TYPE_DOCUMENT,
   SVG_NAMESPACE,
   type NamespaceURI,
@@ -11,6 +12,7 @@ import {
 } from './constants.ts';
 import type {Window} from './Window.ts';
 import type {Node} from './Node.ts';
+import type {Attr} from './Attr.ts';
 import {getElementsByClassName as findElementsByClassName} from './getElementsByClassName.ts';
 import {Event} from './Event.ts';
 import {ParentNode} from './ParentNode.ts';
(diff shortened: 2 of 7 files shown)


andrewiggins:
Closing this PR in favor of two focused replacements:

- **#723 — Add indexed and named NamedNodeMap access:** rebuilt on current `main`, preserving its mutation behavior and including updated benchmarks and WPT classifications.
- **#724 — Detach adopted attributes from their elements:** stacked on #682, with ownership committed before notifications and regression coverage for reentrant adoption.

The remaining mutation fixes are covered by `main`, while general subtree adoption belongs to #682.

PR #713, opened 2026-09-16
Add optional DOM receiver configuration

MitchLillie:
## What this changes

Adds optional per-element member configuration and `blockedProperties` to `DOMRemoteReceiver`. `RemoteReceiverElement` exposes the same options through static host configuration. The options type is exported as `DOMRemoteReceiverOptions`.

```ts
const receiver = new DOMRemoteReceiver({
  elements: {
    'ui-button': {
      properties: {
        disabled: {type: 'boolean'},
        label: {type: 'string', attribute: 'accessible-label'},
        href: {type: 'string', attribute: true},
      },
      attributes: ['slot'],
      events: {click: {}},
      methods: ['focus'],
    },
  },
});
```

Optional types validate non-nullish property values without coercion. Supported types are `string`, `number`, `boolean`, `object`, `array`, and `function`; object and array checks are shallow. Nullish property values are assigned unchanged.

Attribute definitions select permitted names, not automatic reflection or conversion. Attribute values remain strings, with nullish values removing the attribute. The separate `attributes` list can add attribute-only names. Events use DOM names such as `click`.

Existing element-name arrays retain ordinary member handling. Member maps are optional, and the existing `call` callback remains available for custom dispatch. Examples and API documentation cover both configuration styles.

## Compatibility

This includes a **minor** `@remote-dom/core` changeset. There is no blanket migration to required per-element schemas. Data-only receivers are unchanged. The changeset lists the default member exclusions, URL-value checks, and method-dispatch restrictions.

This release also fixes security vulnerabilities.

## Validation

- 746 unit tests passed (`pnpm exec vitest run --maxWorkers=2`).
- 42 browser tests passed, plus two local example smoke checks.
- Type checking, build, formatting, and bundle-size checks passed.

The core host bundle measures 3.17 KB with the additional configuration, checks, and separated validation helpers. This PR moves its budget from 2.5 KB to 3.25 KB; other budgets are unchanged.


MitchLillie:
## Manual verification

I verified this locally using the custom-element example:

```sh
pnpm example:custom-element --host 127.0.0.1 --port 8188 --strictPort
```

Opened http://127.0.0.1:8188 and confirmed:

- **Existing example behavior:** Clicking the button updated its count each time. After the fifth click, the `primary` attribute was removed and the button’s appearance changed accordingly. This used the existing `elements: ['ui-button']` configuration without a member map.
- **Ordinary properties:** Initial `label` and `disabled` values were retained, and a subsequent `label` update was applied. These checks inspected the element’s properties, not its visual state.
- **Configuration checks:** Element selection, default member checks, and additional host `blockedProperties` exclusions behaved as expected. Rejected operations left the rendered content unchanged.
- **URL updates:** A relative URL was accepted. Unsupported values were rejected through both the property and attribute channels, leaving the previously accepted URL unchanged.
- **Optional member lists:** With `properties: ['disabled', 'label']` and `methods: ['focus']`, the label update and focus call succeeded. An additional property and a `blur` call were rejected because they were outside the supplied lists.

These checks covered both the existing element-name array and the optional per-element member configuration.

README.md
```diff
@@ -235,12 +235,9 @@ Finally, we need to provide a “real” implementation of our `ui-button` eleme
...
       class UIButton extends HTMLElement {
-        // By default, `DOMRemoteReceiver` will assign remote properties as properties,
-        // but only if the element has a matching property defined. Otherwise, the remote
-        // properties will be set as attributes. We’ll observe the `primary` attribute
-        // in order to update our rendered content when that attribute changes. We’ll
-        // define an `onClick` method, though, which will be set to the value of the `onClick`
-        // remote property.
+        // The host policy below allows the remote to update `primary` as an
+        // attribute. Property, attribute, and event permissions are separate;
+        // remote properties are not automatically converted to attributes.
```
vividviolet: This seems like a breaking change?

packages/core/source/receivers/DOMRemoteReceiver.ts
```diff
@@ -102,16 +172,35 @@ export class DOMRemoteReceiver {
             ? this.root
             : attached.get(id)!;
 
-        return call
-          ? call(element as any, method, ...args)
-          : (element as any)[method](...args);
+        if (!element || element.nodeType !== NODE_TYPE_ELEMENT) {
+          throw new Error(`Method target is not allowed: ${id}`);
```
vividviolet: nit: can be updated to be more accurate? We're blocking the method on non-elements and for elements that are not attached?

packages/core/source/receivers/DOMRemoteReceiver.ts
```diff
@@ -16,6 +16,33 @@ const REMOTE_IDS = new WeakMap<Node, string>();
...
+  'innerhtml',
+  'outerhtml',
+  'srcdoc',
+  '__proto__',
+  'constructor',
+  'prototype',
+  'is',
+]);
+
+const URL_PROPERTIES =
```
vividviolet, line 125: ### 🔴 `protocol` bypasses the script URL check

This is still exploitable when a host allows `<a>` elements with default properties:

```ts
{href: 'custom:window.__escaped = true; void 0', protocol: 'javascript:'}
```

`href` passes because `custom:` is not a script scheme. Setting `HTMLAnchorElement.protocol` then changes it to `javascript:`. I confirmed script execution in Chromium for both initial properties and a later `UPDATE_PROPERTY`. The URL tests only cover names in `URL_PROPERTIES`, so they stay green.

Fix: Apply `SCRIPT_URL` validation to `protocol` on native `a`/`area` elements. Add initial and update Playwright cases that click the anchor.

packages/core/source/receivers/DOMRemoteReceiver.ts
```diff
@@ -150,11 +240,124 @@ export class DOMRemoteReceiver {
...
       },
     });
 
+    function assertPropertyAllowed(
+      policy: ElementPolicy | undefined,
+      property: string,
+      value: unknown,
+      type: number,
+    ) {
+      const allowed =
```
vividviolet: ### 🟠 `assertPropertyAllowed` is doing too much

It selects the member list, handles an unknown type, checks the name, detects native methods, and validates URL values in one condition. The `protocol` gap above is the kind of case this complexity hides. Can we separate member authorization from value validation?

Fix:

```ts
const allowed = allowedMembersFor(policy, type); // throws for an unknown type
if (allowed && !allowed.has(property)) reject();
if (type === UPDATE_PROPERTY_TYPE_EVENT_LISTENER) return;

const normalized = property.toLowerCase();
if (
  isBlockedProperty(normalized, property, type) ||
  isUnsafeUrlValue(policy, normalized, value, type)
) {
  reject();
}
```

packages/core/source/receivers/DOMRemoteReceiver.ts
```diff
@@ -102,16 +172,35 @@ export class DOMRemoteReceiver {
...
+        }
+
+        const call = options.call;
+        if (call) return call(element as Element, method, ...args);
+
+        const policy = nodePolicies.get(element);
+        if (
+          !policy ||
+          (policy.methods && !policy.methods.has(method)) ||
+          ((!policy.element.includes('-') || method in HTMLElement.prototype) &&
```
vividviolet: Can we move this to `isMethodAllowed()`? The current negative condition combines the optional allowlist with the default native/custom-element policy, so it takes a while to work out what is allowed.

Fix:

```ts
function isMethodAllowed(policy: ElementPolicy, method: string) {
  if (policy.methods && !policy.methods.has(method)) return false;

  return (
    method === 'focus' ||
    method === 'blur' ||
    (policy.element.includes('-') && !(method in HTMLElement.prototype))
  );
}

if (!policy || !isMethodAllowed(policy, method)) {
  throw new Error(`Method is not allowed: ${method}`);
}
```

packages/core/source/receivers/DOMRemoteReceiver.ts
```diff
@@ -41,59 +68,102 @@ export class DOMRemoteReceiver {
...
-     */
-    call?(element: Element, method: string, ...args: any[]): any;
-
-    /**
-     * Controls how DOM elements created in based on remote elements are retained
-     * once they are disconnected from the remote environment.
-     */
-    cache?: {
+  constructor(
+    options: RemoteReceiverOptions & {
```
vividviolet: Can we name this `DOMRemoteReceiverOptions`? The constructor has about 65 lines of inline type documentation, and `RemoteReceiverElement` has to reconstruct the type with `ConstructorParameters<typeof DOMRemoteReceiver>[0]`.

Fix:

```ts
export interface DOMRemoteReceiverOptions extends RemoteReceiverOptions {
  root?: Element;
  elements?: readonly string[] | Readonly<Record<string, DOMRemoteElementPolicy>>;
  blockedProperties?: readonly string[];
  call?(element: Element, method: string, ...args: any[]): any;
  cache?: {maxAge?: number};
}

constructor(options: DOMRemoteReceiverOptions = {}) {
```

.changeset/strict-dom-receiver-policy.md
```diff
@@ -0,0 +1,9 @@
+---
+'@remote-dom/core': minor
+---
+
```
vividviolet, line 4: Can we include a note about how we've fixed some security vulnerability by blocking dangerous props by default and list these out?

packages/core/source/receivers/DOMRemoteReceiver.ts
andrewiggins: Confirmed this against the current PR head (6595c0b). One important detail for the fix: the browser accepts `protocol = 'javascript'` without a trailing colon:

 ```ts
   properties: {
     href: 'custom:window.__escaped = true; void 0',
     protocol: 'javascript',
   }
 ```

This produces a `javascript:` URL and executes on click. It reproduces on both `<a>` and `<area>`, through both initial properties and subsequent property updates. Non-string values such as `['javascript:']` also work through browser coercion.

Consequently, adding `protocol` to `URL_PROPERTIES` and applying the existing `SCRIPT_URL` regex would still leave the colonless variant open. We should either block this setter or validate its scheme-specific semantics, including rejecting non-string values before assignment.

Please include browser regression cases for both elements, initial/update paths, and the colonless/coercion variants.

.changeset/strict-dom-receiver-policy.md
MitchLillie: I added a general comment, not sure we should list in great detail

packages/core/source/receivers/DOMRemoteReceiver.ts
```diff
@@ -41,59 +153,57 @@ export class DOMRemoteReceiver {
...
-      maxAge?: number;
-    };
-  } = {}) {
+  constructor(options: DOMRemoteReceiverOptions = {}) {
+    const {root, elements, retain, release, cache} = options;
     this.root = root ?? document.createDocumentFragment();
 
     const {attached} = this;
     const destroyTimeouts = new Map<string, number>();
+    const policies = new Map<string, ElementPolicy>();
```
vividviolet: ### Move policy compilation out of the constructor

This block only snapshots host configuration and creates a lookup. It does not use receiver state. Can we move it out so the constructor only owns the connection lifecycle?

Fix:

```ts
const policyFor = createElementPolicyResolver(elements);
```

```ts
function createElementPolicyResolver(
  elements: DOMRemoteReceiverOptions['elements'],
) {
  const policies = new Map<string, ElementPolicy>();
  const entries: [string, DOMRemoteElementPolicy][] = Array.isArray(elements)
    ? elements.map((name) => [name, {}])
    : Object.entries(elements ?? {});

  for (const [name, definition] of entries) {
    let properties: Map<string, DOMRemotePropertyPolicy['type']> | undefined;
    let attributes =
      definition.attributes && new Set(definition.attributes);

    if (definition.properties) {
      properties = new Map();
      attributes ??= new Set();

      for (const [property, {type, attribute = true}] of Object.entries(
        definition.properties,
      )) {
        properties.set(property, type);

        if (attribute !== false) {
          attributes.add(
            typeof attribute === 'string'
              ? attribute
              : property.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase(),
          );
        }
      }
    }

    policies.set(name, {
      element: name.toLowerCase(),
      properties,
      attributes,
      events: definition.events && new Set(Object.keys(definition.events)),
      methods: definition.methods && new Set(definition.methods),
    });
  }

  return function policyFor(name: string) {
    let policy = policies.get(name);

    if (!policy && elements === undefined) {
      policy = {element: name.toLowerCase()};
      policies.set(name, policy);
    }

    return policy;
  };
}
```

packages/core/source/receivers/DOMRemoteReceiver.ts
```diff
@@ -150,11 +276,121 @@ export class DOMRemoteReceiver {
...
+          (text.nodeType !== NODE_TYPE_TEXT &&
+            text.nodeType !== NODE_TYPE_COMMENT)
+        ) {
+          throw new Error(`Text update target is not allowed: ${id}`);
+        }
+        (text as Text | Comment).data = newText;
       },
     });
 
+    function assertPropertyAllowed(
```
vividviolet: ### Hoist the configured property validator

This function only stays inside the constructor to capture `blockedProperties`. Can we build the validator once and keep all member-policy helpers together at module scope?

Fix:

```ts
const assertPropertyAllowed = createPropertyValidator(
  options.blockedProperties,
);
```

```ts
function createPropertyValidator(
  additionalBlockedProperties: readonly string[] = [],
) {
  const blockedProperties = new Set(BLOCKED_PROPERTIES);

  for (const property of additionalBlockedProperties) {
    blockedProperties.add(property.toLowerCase());
  }

  return function assertPropertyAllowed(
    policy: ElementPolicy | undefined,
    property: string,
    value: unknown,
    type: number,
  ) {
    if (typeof property !== 'string') {
      throw new Error('Member name is not allowed');
    }
    if (!policy) rejectProperty(property, type);

    const allowed = allowedMembersFor(policy, type);
    if (allowed && !allowed.has(property)) rejectProperty(property, type);
    if (type === UPDATE_PROPERTY_TYPE_EVENT_LISTENER) return;

    const normalized = property.toLowerCase();
    if (
      isBlockedProperty(
        policy,
        property,
        normalized,
        type,
        blockedProperties,
      ) ||
      !isValueAllowed(policy, property, value, type) ||
      isUnsafeUrlValue(policy, normalized, value, type)
    ) {
      rejectProperty(property, type);
    }
  };
}

function isBlockedProperty(
  policy: ElementPolicy,
  property: string,
  normalized: string,
  type: number,
  blockedProperties: ReadonlySet<string>,
) {
  return (
    blockedProperties.has(normalized) ||
    normalized.startsWith('on') ||
    (type === UPDATE_PROPERTY_TYPE_PROPERTY &&
      ((property === 'protocol' &&
        (policy.element === 'a' || policy.element === 'area')) ||
        isNativeMethod(property)))
  );
}
```

vividviolet (commented):
Code looks good but I have 2 clean up suggestions to move stuff into pure functions so we can simplify the constructor to make it more readable

packages/core/source/receivers/DOMRemoteReceiver.ts
MitchLillie: Done! 

Guard: [DOMRemoteReceiver.ts:544–547](https://github.com/Shopify/remote-dom/blob/5ba6054/packages/core/source/receivers/DOMRemoteReceiver.ts#L544-L547)
Browser regressions: [dom-receiver-protocol.e2e.ts:7–13](https://github.com/Shopify/remote-dom/blob/5ba6054/e2e/dom-receiver-protocol.e2e.ts#L7-L13)

vividviolet (approved):
Looks good! Thanks for bearing through multiple rounds of feedback

andrewiggins (approved):

